PR #28, opened 2024-08-19
chore: CP-8911 update packages

vvava:
## Description

We had an issue with the tokens' price when a bridging tx  was in progress. The price sometimes was 0. 

## Changes

The bridge config did not be loaded therefor it threw an error which was an empty object. Now we are waiting for the service to get the config before we want to get the data from there.
https://github.com/ava-labs/avalanche-sdks/pull/664

## Testing

Try a bridging tx and check the price of the token. 

## Checklist for the author


meeh0w:
@vvava I tested and it works:
image

I think we can close this PR like you said yesterday - there is another one in queue which updates to a higher version :)

vvava:
duplicated by #28

PR #37, opened 2024-08-30
fix: treat BTC as known network

meeh0w:
## Description
* [CP-9154](https://ava-labs.atlassian.net/browse/CP-9154)

## Changes
* When network is already supported by the extension, do not use the request's payload, but rather resolve the network object via `NetworkService` and use that instead.

## Testing
1. Open extension, switch to C-Chain, go to Send
2. Connect to Core Web
3. Switch to Bitcoin _**via Core Web**_
4. Open extension again, you should not see `Unknown error`

### Before

https://github.com/user-attachments/assets/e8f54b13-4318-47ec-abf0-f49eaaf2e601

### After

https://github.com/user-attachments/assets/685504c9-72c0-4fe7-b124-b0e0f24ba267

## Checklist for the author

### DIFF
--- a/src/background/services/network/handlers/wallet_addEthereumChain.ts
+++ b/src/background/services/network/handlers/wallet_addEthereumChain.ts
@@ -188,7 +188,7 @@ export class WalletAddEthereumChainHandler extends DAppRequestHandler {
       await this.networkService.saveCustomNetwork(network);
     }
 
-    await this.networkService.setNetwork(domain, network.chainId);
+    await this.networkService.setNetwork(domain, network.caipId);
   }
 
   onActionApproved = async (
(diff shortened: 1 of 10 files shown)


src/background/services/network/handlers/wallet_addEthereumChain.ts
```diff
@@ -118,8 +118,14 @@ export class WalletAddEthereumChainHandler extends DAppRequestHandler {
     }
     const skipApproval = await isCoreWeb(request);
 
+    const targetNetwork = chainRequestedIsSupported
```
gergelylovas, line 121: Should this be handled in the `setNetwork` function of the NetworkService instead? Ensuring that we are using the exact network object from the network service appears to be quite error prone.

gergelylovas (changes requested):

src/background/services/network/handlers/wallet_addEthereumChain.ts
meeh0w: I updated `NetworkService.setNetwork()` to accept the chain id instead, this guarantees the network needs to exist in the `NetworkService` first (so dapps / handlers need to first `.saveCustomNetwork()` and then try to switch to it, otherwise it won't work - previously it would).

gergelylovas (approved):

vvava (approved):
seems working well 👍

PR #38, opened 2024-09-04
feat: account renaming via rpc

meeh0w:
## Description
* [CP-8963](https://ava-labs.atlassian.net/browse/CP-8963)
* Related PR: https://github.com/ava-labs/extension-avalanche-playground/pull/50

## Changes
* Adds `avalanche_renameAccount` request handler
* Migrates UI to use the new handler instead of `ExtensionRequest.ACCOUNT_RENAME` and removes the old handler
* **Maintenance:** _replaces `isCoreWeb()` with `isSyncDomain()`, as the latter also supports feature branch deploys of Core Web (e.g. `develop.core-web.pages.dev)`_

## Testing
* Use playground to rename the account (see PR link above)
* Rename account in the extension itself

## Screenshots:
_N/A_

## Checklist for the author


src/background/services/accounts/handlers/avalanche_renameAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+    const account = this.accountsService.getAccountByID(accountId);
+
+    if (!account) {
+      return {
+        ...request,
+        error: ethErrors.rpc.invalidParams({
+          message: 'No account with specified ID',
+        }),
+      };
+    }
```
bferenc, line 54: Can we also validate the `newName` param? Currently, the handler accepts empty string.
Would be great if we added the same validation to our own UI as well (it also accepts empty strings)

src/background/services/accounts/handlers/avalanche_renameAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (isSyncDomain(request.site.domain)) {
```
bferenc, line 56: Should we add a check to `isSyncDomain` so it returns `true` only if the request is coming from the `active` tab?
This way synced domains won't be able to make unnoticed changes in the users' wallets.

bferenc (changes requested):
Nice! Some questions below 👇

src/background/services/accounts/handlers/avalanche_renameAccount.ts
meeh0w: Sure thing :) 6a2d3dded

src/background/services/accounts/handlers/avalanche_renameAccount.ts
meeh0w: Good point! 0ab8239b6

bferenc (dismissed):
💯

meeh0w:
@bferenc I had to resolve conflicts 🙏

bferenc (approved):

vvava:
Don't we want to accept account indexes as well? I mean not just account ids.

vvava (approved):

PR #50, opened 2024-09-18
feat: remove walletService from accountsService

vvava:
## Checklist for the author

### DIFF
--- a/src/background/services/accounts/AccountsService.ts
+++ b/src/background/services/accounts/AccountsService.ts
@@ -15,7 +15,6 @@ import {
   WalletId,
 } from './models';
 import { OnLock, OnUnlock } from '@src/background/runtime/lifecycleCallbacks';
-import { WalletService } from '../wallet/WalletService';
 import { NetworkService } from '../network/NetworkService';
 import { NetworkVMType } from '@avalabs/core-chains-sdk';
 import { PermissionsService } from '../permissions/PermissionsService';
@@ -24,6 +23,9 @@ import { DerivedAddresses, SecretType } from '../secrets/models';
 import { isPrimaryAccount } from './utils/typeGuards';
 import { AnalyticsServicePosthog } from '../analytics/AnalyticsServicePosthog';
 import getAllAddressesForAccount from '@src/utils/getAllAddressesForAccount';
+import { SecretsService } from '../secrets/SecretsService';
+import { LedgerService } from '../ledger/LedgerService';
+import { WalletConnectService } from '../walletConnect/WalletConnectService';
 
 type AddAccountParams = {
   walletId: string;
@@ -82,10 +84,12 @@ export class AccountsService implements OnLock, OnUnlock {
 
   constructor(
     private storageService: StorageService,
-    private walletService: WalletService,
     private networkService: NetworkService,
     private permissionsService: PermissionsService,
-    private analyticsServicePosthog: AnalyticsServicePosthog
+    private analyticsServicePosthog: AnalyticsServicePosthog,
+    private secretsService: SecretsService,
+    private ledgerService: LedgerService,
+    private walletConnectService: WalletConnectService
   ) {}
 
   async onUnlock(): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/background/services/accounts/AccountsService.ts
```diff
@@ -82,10 +84,12 @@ export class AccountsService implements OnLock, OnUnlock {
...
   constructor(
     private storageService: StorageService,
-    private walletService: WalletService,
     private networkService: NetworkService,
     private permissionsService: PermissionsService,
-    private analyticsServicePosthog: AnalyticsServicePosthog
+    private analyticsServicePosthog: AnalyticsServicePosthog,
+    private secretsService: SecretsService,
+    private ledgerService: LedgerService,
+    private walletConnectService: WalletConnectService
```
meeh0w, line 92: Should they be dependencies of the `SecretsService` instead? If I see correctly, there should be no circular dependencies and the only reason we inject them here is to pass them down to the `SecretsService` calls later on.

src/background/services/secrets/SecretsService.ts
```diff
@@ -209,17 +237,14 @@ export class SecretsService {
     throw new Error('Unsupported import type');
   }
 
-  async getActiveAccountSecrets() {
+  async getActiveAccountSecrets(activeAccount: Account) {
```
meeh0w, line 240: Since we pass the account as a parameter, I think it would make sense to change this method's signature to `getAccountSecrets(account: Account)`. It no longer matters if it is the active account or no :)

src/background/services/secrets/SecretsService.ts
```diff
@@ -146,23 +167,30 @@ export class SecretsService {
...
-  getActiveWalletSecrets(walletKeys: WalletSecretInStorage) {
-    const accountsService = container.resolve(AccountsService);
-
-    const activeWalletId = isPrimaryAccount(accountsService.activeAccount)
-      ? accountsService.activeAccount.walletId
-      : accountsService.activeAccount?.id;
+  getActiveWalletSecrets(
+    walletKeys: WalletSecretInStorage,
+    activeAccount: Account
+  ) {
```
meeh0w, line 173: Similar comment to the one about `getActiveAccountSecrets` method - this method is more generic now, we can pass any account and therefore retrieve any wallet's secrets:

```suggestion
  getWalletSecretsForAccount(
    walletKeys: WalletSecretInStorage,
    account: Account
  ) {
```

src/background/services/secrets/SecretsService.ts
```diff
@@ -257,7 +282,8 @@ export class SecretsService {
   }
 
   async deleteImportedWallets(
-    ids: string[]
+    ids: string[],
+    walletConnectService: WalletConnectService
```
meeh0w, line 286: Can we inject `WalletConnectService` in the constructor instead of passing it through the method parameters here?

src/background/services/secrets/SecretsService.ts
```diff
@@ -477,4 +522,291 @@ export class SecretsService {
...
+    }
+
+    return addresses;
+  }
+
+  async addAddress({
+    index,
+    walletId,
+    ledgerService,
+    networkService,
```
meeh0w, line 629: Can we inject `LedgerService` and `NetworkService` in the constructor instead of passing them through the method parameters here?

meeh0w (changes requested):
Nicely done! I left some small comments, but great work overall 👏

src/background/services/accounts/AccountsService.ts
vvava: The reason is why the dependencies got out of the SecretsService and took place in the AccountsService, because we talked about these cases with Gergo and the goal would be the SecretsService should have only one dependency in general, the StorageService, because it communicates basically only that service and the others are used in special cases in "its own" individual functions so the SecretService does not need those services in general. I know they are just moved to an other file in this case but it seems to be more rational.

meeh0w: Makes sense 👍

meeh0w (approved):

bferenc (approved):

PR #59, opened 2024-10-04
feat: allow getAddressesInRange for non-core dApps

meeh0w:
## Description
* [CP-9260](https://ava-labs.atlassian.net/browse/CP-9260)

## Changes
* Allows `avalanche_getAddressesInRange` to be called by non-Core dApps, but shows an approval window then

## Testing
* via playground

## Screenshots:

https://github.com/user-attachments/assets/baddc239-e2f1-4605-80a9-cca9cd73afc7

## Checklist for the author


gergelylovas (approved):

gergelylovas:
Tested as well, works nicely!

vvava (approved):
We have a bit strange scrolling behaviour but I think it is a more general problem so we can address it in another ticket?

https://github.com/user-attachments/assets/79ffe0b3-6a52-401d-8963-c74cfd50831b

gergelylovas (approved):

meeh0w:
@vvava Yup, good catch. I noticed the scrollbars issue usually appears when we have `<Scrollbars>` with `<Tooltip>` inside, but I did not find a quick solution to this yet. We should definitely revisit this on a broader scale at a later point.

vvava (approved):

PR #62, opened 2024-10-04
feat: CP-9258 add new calculation

vvava:
## Description

We want to use the fee calculation from the vm-modules package instead of doing it in-place.

WIP: test cases

## Changes

Integrate the `getNetworkFee` from vm-modules.

## Testing

Go to a transaction in an EVM network -> check the fees displayed correctly.

## Checklist for the author


src/background/services/networkFee/NetworkFeeService.test.ts
```diff
@@ -54,37 +55,38 @@ describe('src/background/services/networkFee/NetworkFeeService', () => {
...
-        },
-        medium: {
-          maxFee: maxFeePerGas + BigInt(2e9),
-          maxTip: BigInt(2e9),
-        },
-        high: {
-          maxFee: maxFeePerGas + BigInt(3e9),
-          maxTip: BigInt(3e9),
-        },
+    it.only('provides base information and three rate presets', async () => {
```
meeh0w, line 66:
```suggestion
    it('provides base information and three rate presets', async () => {
```

src/background/services/networkFee/NetworkFeeService.ts
```diff
@@ -1,46 +1,35 @@
...
-  medium: 4n, // 4x base
-  high: 6n, // 6x base
-};
-
 @singleton()
 export class NetworkFeeService {
   constructor(private moduleManager: ModuleManager) {}
 
   async getNetworkFee(network: NetworkWithCaipId): Promise<NetworkFee | null> {
+    const module = await this.moduleManager.loadModuleByNetwork(network);
```
meeh0w, line 14: I think these `vmName` conditions can be removed entirely. `.getNetworkFee()` response has the same shape for all VM modules.

meeh0w (changes requested):
Looks great! I left just a small comment, I think we can simplify it even more :)

meeh0w (approved):

bferenc (approved):

PR #63, opened 2024-10-08
fix: CP-9331 add avacloud to whitelist

vvava:
## Description

 AvaCloud users have to approve the expose addresses transaction after browser refreshing

## Changes

AvaCloud has been added to the whitelist besides Core web

## Testing

Go to avacloud -> try to reproduce the ticket description [CP-9331](https://ava-labs.atlassian.net/jira/software/c/projects/CP/boards/72?selectedIssue=CP-9331)

## Checklist for the author

### DIFF
--- a/src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
+++ b/src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
@@ -22,6 +22,12 @@ type Params = [
 ];
 import { AccountsService } from '../AccountsService';
 
+const EXPOSED_DOMAINS = [
+  'develop.avacloud-app.pages.dev',
+  'avacloud.io',
+  'staging--ava-cloud.avacloud-app.pages.dev',
+];
+
 @injectable()
 export class AvalancheGetAddressesInRangeHandler extends DAppRequestHandler<
   Params,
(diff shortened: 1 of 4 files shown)


gergelylovas (dismissed):

src/background/services/network/utils/getSyncDomain.ts
```diff
@@ -5,6 +5,9 @@ const SYNCED_DOMAINS = [
   'core.app',
   'test.core.app',
   runtime.id,
+  'develop.avacloud-app.pages.dev',
+  'avacloud.io',
+  'staging--ava-cloud.avacloud-app.pages.dev',
```
gergelylovas, line 10: I think we should have a separate list of the getAddressesInRange list, since we don't want to give the same ability for all RPCs for AvaCloud. Like account renaming, deletion or creation should not be exposed to them. Also this way if the network is set to BTC in Core, Avacloud would need to switch to BTC as well, which caused issues in the past.

gergelylovas (changes requested):

src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
```diff
@@ -140,7 +146,13 @@ export class AvalancheGetAddressesInRangeHandler extends DAppRequestHandler<
         externalLimit: correctedExternalLimit,
       });
 
-      if (await canSkipApproval(request.site.domain, request.site.tabId)) {
+      if (
+        await canSkipApproval(
```
gergelylovas, line 150: Can you please update the tests of this file to check for this change as well?

gergelylovas (approved):

meeh0w (approved):

bferenc (approved):

PR #65, opened 2024-10-11
fix: bridge issues

meeh0w:
* Fixes for bridging issues found during blue build regression tests


vvava (approved):

bferenc (approved):
Tested bridging initiated from within the extension and on core web, both worked 👍

PR #67, opened 2024-10-14
fix: getAddresses popup appearing randomly on core.app

meeh0w:
Fix for `Expose addresses?` popup showing up randomly on core.app.

It seems core.app is polling for accounts and reacting to accounts being created/renamed in the extension. If core.app is not the active tab, we show up the approval window.

### DIFF
--- a/src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
+++ b/src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
@@ -147,11 +147,10 @@ export class AvalancheGetAddressesInRangeHandler extends DAppRequestHandler<
       });
 
       if (
-        await canSkipApproval(
-          request.site.domain,
-          request.site.tabId,
-          EXPOSED_DOMAINS
-        )
+        await canSkipApproval(request.site.domain, request.site.tabId, {
+          domainWhitelist: EXPOSED_DOMAINS,
+          allowInactiveTabs: true,
+        })
       ) {
         return {
           ...request,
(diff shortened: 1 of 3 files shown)


src/utils/canSkipApproval.ts
```diff
@@ -3,14 +3,23 @@ import { isSyncDomain } from '@src/background/services/network/utils/getSyncDoma
 import { isActiveTab } from './isActiveTab';
 import { runtime } from 'webextension-polyfill';
 
+type SkipApprovalOptions = {
+  allowInactiveTabs?: boolean;
```
meeh0w, line 7: I went with dev having to explicitly say `allowInactiveTabs` rather than making the `tabId` optional. It feels less error-prone this way.

src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
```diff
@@ -147,11 +147,10 @@ export class AvalancheGetAddressesInRangeHandler extends DAppRequestHandler<
...
       if (
-        await canSkipApproval(
-          request.site.domain,
-          request.site.tabId,
-          EXPOSED_DOMAINS
-        )
+        await canSkipApproval(request.site.domain, request.site.tabId, {
+          domainWhitelist: EXPOSED_DOMAINS,
+          allowInactiveTabs: true,
+        })
```
vvava, line 153: Should we put all of the parameters to an object then?

vvava (approved):

ryanewood (approved):
I haven't been able to reproduce the approval address screen appearing randomly using this branch like I have on blue build. lgtm 👍

src/background/services/accounts/handlers/avalanche_getAddressesInRange.ts
meeh0w: I can simplify in a follow up PR (so I don't need to collect approvals after updating again 😄)

PR #73, opened 2024-10-18
fix: collectibles view quirks

meeh0w:
_no ticket, fixes for issues found during blue build testing (+ some additional quirks)_

## Fixed issues
* the weird UI issue where ERC-1155 balances would show up bunched together before the images load
* toggling visibility of NFTs (previously we would toggle based on NFT's `address` only, but it's possible to have multiple NFTs of the same address but different `tokenId` -- this resulted in hiding entire batches of NFTs when trying to hide a specific NFT)
* the Collectibles tab appearing in grid view for a fraction of a second even if the user switched to list view (extension takes a moment to load the preferred state from storage)

## Testing
Go to the Collectibles tab and play around

## Checklist for the author

### DIFF
--- a/src/contexts/SettingsProvider.tsx
+++ b/src/contexts/SettingsProvider.tsx
@@ -27,6 +27,7 @@ import {
   SetStateAction,
 } from 'react';
 import { filter, map } from 'rxjs';
+import { omit, set } from 'lodash';
 import { useConnectionContext } from './ConnectionProvider';
 import { getCurrencyFormatter } from './utils/getCurrencyFormatter';
 import { updateIfDifferent } from '@src/utils/updateIfDifferent';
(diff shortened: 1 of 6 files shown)


src/contexts/SettingsProvider.tsx
```diff
@@ -146,29 +147,28 @@ export function SettingsContextProvider({ children }: { children: any }) {
   );
 
   async function toggleCollectibleVisibility(nft: NftTokenWithBalance) {
-    const key = nft.address;
-    const collectiblesVisibility = settings?.collectiblesVisibility ?? {};
+    const key = `${nft.address}-${nft.tokenId}`;
+    const visibility = settings?.collectiblesVisibility ?? {};
+    // We used to (wrongly) index by address only.
+    const isHidden = (visibility[key] ?? visibility[nft.address]) === false;
```
meeh0w, line 153: Having a schema upgrade would be the best for this, but I don't think it's worth the effort (+ we don't have the `tokenId` in the current storage data anyway). After toggling once it "fixes itself" for all NFTs sharing a given `address`.

bferenc (approved):

zsuninnyi (approved):

vvava (approved):
the previous one was a wrong account :D

PR #76, opened 2024-10-21
fix: ledger btc policy issues

meeh0w:
* [CP-9365](https://ava-labs.atlassian.net/browse/CP-9365)

## Changes
Two issues fixed:
* the `account` parameter for `SecretsService.getBtcWalletPolicyDetails` method was optional, and we were not providing it in some cases, resulting in the method returning `undefined` (UI did not know the policy was already stored)
* `LedgerProvider` was spamming the device with requests, because some functions were not memoized (they were changing, causing `useEffect()` to run and query the device for `masterFingerprint`). Those requests sometimes coincided with actual sign requests, causing them to fail with `Action was already pending on the device` error

## Testing
* Onboard with Ledger
* Sign a BTC transaction (you should be prompted to store the BTC policy)
* Sign another BTC transaction from the same account (you should NOT be prompted to store the BTC policy)

## Checklist for the author

### DIFF
--- a/src/background/services/secrets/SecretsService.test.ts
+++ b/src/background/services/secrets/SecretsService.test.ts
@@ -883,7 +883,7 @@ describe('src/background/services/secrets/SecretsService.ts', () => {
       mockMnemonicWallet();
 
       await expect(
-        secretsService.getBtcWalletPolicyDetails()
+        secretsService.getBtcWalletPolicyDetails(activeAccountData)
       ).resolves.toBeUndefined();
     });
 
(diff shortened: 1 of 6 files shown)


vvava (approved):

bferenc (changes requested):
Wallet policy registration shows up only once and I can sign any number of transactions with it as expected 👍 
However:
- I'm expected to do it again if I switch accounts even though I'm using BIP44 (that should use the xpub only once for all accounts I believe?)
- after switching account and confirming the registration attempt on the device I get an "already stored" error and can't proceed

bferenc (approved):
Actually it was an issue on my end, works as expected. Sorry 🙃

PR #82, opened 2024-10-25
feat: migrate to Unified Bridge, introduce the ICTT bridge

meeh0w:
* [CP-9105](https://ava-labs.atlassian.net/browse/CP-9105)

## Description
* 👋 Bye bye `@avalabs/bridge-sdk` _(in the frontend -- some backend parts need to stay for a while)_
* 🤗 Embrace the `@avalabs/bridge-unified` _(comes in with CTTP, ICTT (erc20-erc20) and Avalanche Bridge integrations)_

## Changes
* Lots of code removed
    * Some parts that are still using the `@avalabs/bridge-sdk` are still left - during the migration phase, some people may have old `BridgeTransaction`s either in progress or completed, but not acknowledged in the UI yet. We may need a schema migration, depending on how the UnifiedBridgeSDK handles BTC tracking.
* Simplified data flows:
    * No more `useAvalancheBridge()` / `useBtcBridge()` / `useEthBridge()` / `useUnifiedBridge()` hooks. The old `useBridge()` which previously merged the capabilities of them all is now enough to handle everything.
    * No more `BridgeFormETH` / `BridgeFormBTC` / `BridgeFormAVAX` or `BridgeFormUnified`. The `BridgeForm` got promoted and is used directly by `Bridge.tsx` page.
    * Some unused components & utils got removed

## The new user path
1. User selects assets from those available for bridging on the current active network
2. Target chains are loaded after asset is selected. The first available target chain is pre-selected.
    * User may change the target chain if there are more options than just one.
3. User may change the source network at any point (either by changing it in the form, or via the networks widget)
    * If asset is present on the new network, the token & amount is kept in-tact.
    * If asset is not present on the new network, form is reset.

## Testing
* Entire Bridge suite must be tested.
* Remember about testing the Core Web integration as well.
* Test how it behaves when the related feature flags are toggled:
    * `unified-bridge-cctp`
    * `unified-bridge-ictt`
    * `unified-bridge-ab-evm`
    * `unified-bridge-ab-ava-to-btc`
    * `unified-bridge-ab-btc-to-ava`

### Feature Flags examples
_When Bitcoin -> Avalanche bridge **IS** disabled_
image

_The moment when selected bridge route (Avalanche -> Bitcoin) **gets** disabled_

https://github.com/user-attachments/assets/8acce9cf-8d31-4458-99d4-88a2a1123e24

## Checklist for the author


src/pages/Bridge/utils/findMatchingBridgeAsset.ts
```diff
@@ -0,0 +1,27 @@
...
+  TokenWithBalance,
+} from '@avalabs/vm-module-types';
+
+export const findMatchingBridgeAsset = (
+  assets: BridgeAsset[],
+  token: Exclude<TokenWithBalance, NftTokenWithBalance>
+): BridgeAsset | undefined => {
+  return assets.find((a) => {
+    if (a.type === BridgeTokenType.NATIVE && token.type === TokenType.NATIVE) {
+      return a.symbol.toLowerCase() === a.symbol.toLowerCase();
```
gergelylovas, line 18:
```suggestion
      return a.symbol.toLowerCase() === token.symbol.toLowerCase();
```

gergelylovas (changes requested):

src/pages/Bridge/utils/findMatchingBridgeAsset.ts
meeh0w: Good eyes! Fixed in [540a524](https://github.com/ava-labs/core-extension/pull/82/commits/540a5248068a082c2cbf90170d415e5ae968ea32)

meeh0w: I also fixed a few other issues and added missing support for dev environment of the avalanche bridge.

gergelylovas (approved):

bferenc (commented):
Unrelated to this changes, just FYI: the `tokenlist` endpoint responds with an empty array fro sepolia testnet, so it's only possible to use ETH for briding from sepolia to fuji C.

My sepolia -> fuji C ethereum bridge finished successfully, however the tracker just keeps trying 😅 
image

And the BTC bridge transaction is not recognized in the activity log
image

Do we want to address these issues in this PR?

meeh0w:
@bferenc, thanks for testing! 🍻 

* `tokenlist` was this way for a long long time. I agree we could populate it with some tokens to make our lives easier.
* I tried to reproduce the hanging tracking screen, but couldn't (I tried once for CCTP and twice for ETH -> ETH.e).
* I fixed the BTC bridge recognition errors (the `@avalabs/bridge-unified` version update takes care of that)
* I addressed some issues with amount formatting, gas calculation (bridge amount was always taken into consideration there, even when bridged asset was not the gas token) and with some values not being reset when token was being changed (i.e. receive amount would stay populated)

bferenc (approved):

gergelylovas (approved):

PR #83, opened 2024-10-28
fix: CP-9311 x and p send set address

vvava:
## Description

We don't want to completely control the contact input anymore. 

## Changes

Let the user changes the address to any value.

## Testing

Go to send on X or P chain -> try to change the address -> it hould let you set an empty string or whatever you want meanwhile it validates correctly.

## Screenshots:

https://github.com/user-attachments/assets/67504f22-8190-4d1e-bd88-1e8edf29964c

## Checklist for the author

### DIFF
--- a/src/pages/Send/components/SendAVM.tsx
+++ b/src/pages/Send/components/SendAVM.tsx
@@ -110,7 +110,7 @@ export const SendAVM = ({
       token={nativeToken}
       tokenList={tokenList}
       onContactChanged={(contact) => {
-        if (contact?.addressXP) setAddress(contact.addressXP);
+        setAddress(contact?.addressXP ?? '');
       }}
       onAmountChanged={(newAmount) => setAmount(newAmount)}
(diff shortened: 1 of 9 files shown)


meeh0w (changes requested):
I see a few possible problems here:

* It's no longer possible to enter the address without the prefix, which I think should always be possible.
    * If we want the prefix to be required (and correspond to the active network), we should have a clear error message for it, for example `Invalid address prefix, X-Chain address is required`. Right now it says `Address is invalid`, but it is in fact valid.
* When you enter the address without the prefix, it does not recognize it anymore - even if it's your own account.

https://github.com/user-attachments/assets/92fe1e58-4739-4dd0-818a-ab36f8421bd8

meeh0w (approved):
It still does not recognize unprefixed addresses, but it's not a biggie I think 👍

bferenc (approved):

PR #84, opened 2024-10-28
feat: add default tokens to swap

vvava:
## Description

We want to add default tokens to the Swap page

## Changes

Set the AVAX and USDC tokens as default transaction pairs.

## Testing

Go to the swap page -> you should see the AVAX in the `from` and the USDC in the `to` field -> swap should work as before

## Screenshots:

https://github.com/user-attachments/assets/0cc4998b-7627-4284-ad56-e807c69cd9ba

## Checklist for the author

### DIFF
--- a/src/pages/Swap/Swap.tsx
+++ b/src/pages/Swap/Swap.tsx
@@ -1,6 +1,6 @@
 import { useSwapContext } from '@src/contexts/SwapProvider/SwapProvider';
 import { useTokensWithBalances } from '@src/hooks/useTokensWithBalances';
-import { useEffect, useMemo, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { resolve } from '@src/utils/promiseResolver';
 import { TransactionDetails } from './components/TransactionDetails';
 import { PageTitle } from '@src/components/common/PageTitle';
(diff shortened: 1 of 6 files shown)


src/pages/Swap/Swap.tsx
```diff
@@ -104,6 +105,30 @@ export function Swap() {
     destAmount,
   } = useSwapStateFunctions();
 
+  useEffect(() => {
+    if (!selectedFromToken && !selectedToToken) {
```
meeh0w, line 109: This check alone does not quite work, tokens are not restored when I close and re-open extension.

### Prod build:

https://github.com/user-attachments/assets/11e866de-ffec-499f-8af0-28aa140a065c


### This branch:

https://github.com/user-attachments/assets/b4f6887e-925c-4b59-87f2-b14a2eb94d14

meeh0w (changes requested):

meeh0w (changes requested):
There is still something wrong with remembering the values. It sometimes works and sometimes doesn't. There's probably a race condition somewhere.

https://github.com/user-attachments/assets/24f14d30-a466-4cc4-9480-850f3a19c74b

vvava:
> There is still something wrong with remembering the values. It sometimes works and sometimes doesn't. There's probably a race condition somewhere.
> 
>  Screen.Recording.2024-10-31.at.11.53.19.mov

Yeah, something is wrong there, I'm fixing it

meeh0w (approved):
Looks like it's working now 👏

bferenc (approved):

PR #90, opened 2024-11-14
feat: CP-9472 add duplicated account dialog

vvava:
## Description

We want to notify the user when a private key is being imported and that is already exist as a primary wallet.
Also we want to notify the users in a sooner step when they want to add a ledger wallet which is already lives in the extension.

## Changes

New dialog to notify the users about duplicated account.
Modified import ledger handler so it can be used as a monitoring function to get the wallet is exist.
Modified UI flow to notify the users the ledger wallet already exists before we  moved them forward to the naming phase.

## Testing

Go to the Accounts page -> try to add the existing mnemonic wallet by importing the private key
and check you cannot add the same ledger (with BIP44 and Ledger Live) multiple times.

## Screenshots:

image

https://github.com/user-attachments/assets/72956a20-de26-41b8-9492-25ba796c729c

## Checklist for the author


src/background/services/wallet/handlers/importLedger.ts
```diff
@@ -68,6 +70,16 @@ export class ImportLedgerHandler implements HandlerType {
       };
     }
 
+    if (onlyCheckWalletIsExist) {
+      return {
+        ...request,
+        result: {
+          id: '0',
+          type: secretType,
+        },
```
gergelylovas, line 79: Let's try to preserve the same response type as the success message normally has.
```suggestion
        result: {
          id: '0',
          name,
          type: secretType,
        },
```

src/background/services/wallet/handlers/models.ts
```diff
@@ -23,6 +23,7 @@ export type ImportLedgerWalletParams = {
   pubKeys?: PubKeyType[];
   secretType: SecretType.Ledger | SecretType.LedgerLive;
   name?: string;
+  onlyCheckWalletIsExist?: boolean;
```
gergelylovas, line 26: Should we call this `dryRun` instead? 
```suggestion
  dryRun?: boolean;
```

src/pages/ImportPrivateKey/ImportPrivateKey.tsx
```diff
@@ -49,14 +51,38 @@ export function ImportPrivateKey() {
...
+        ({ addressC }) => addressC.toLowerCase() === address.toLowerCase()
+      );
+      if (findAccount) {
+        setIsKnownAccount(true);
+      }
+    },
+    [allAccounts]
+  );
+
+  console.log('allAccounts: ', allAccounts);
```
gergelylovas, line 73: Please remove

src/components/ledger/LedgerConnector.tsx
```diff
@@ -48,11 +50,13 @@ export interface LedgerConnectorData {
 interface LedgerConnectorProps {
   onSuccess: (data: LedgerConnectorData) => void;
   onTroubleshoot: () => void;
+  checkWalletIsExist?: boolean;
```
gergelylovas, line 53:
```suggestion
  checkWalletExists?: boolean;
```

gergelylovas (changes requested):

bferenc (changes requested):
Works great! 👍 
Can we standardize the names though? Currently there are multiple cases like "checkWalletExists", "isLedgerExists", "checkLedgerWalletIsExist"

bferenc (approved):
Tested as well ✅

src/pages/ImportPrivateKey/ImportPrivateKey.tsx
```diff
@@ -49,14 +51,36 @@ export function ImportPrivateKey() {
   const balance = useBalanceTotalInCurrency(derivedAddresses as Account);
   const { isImporting: isImportLoading, importPrivateKey } =
     usePrivateKeyImport();
+  const { selectAccount } = useAccountsContext();
   const history = useHistory();
+  const { allAccounts } = useAccountsContext();
+  const [isKnownAccount, setIsKnownAccount] = useState(false);
+  const [isDuplicatedAccountDialogOpen, setIsDuplicatedDialogDialogOpen] =
```
meeh0w, line 58:
```suggestion
  const [isDuplicatedAccountDialogOpen, setIsDuplicatedAccountDialogOpen] =
```

src/pages/ImportPrivateKey/ImportPrivateKey.tsx
```diff
@@ -49,14 +51,36 @@ export function ImportPrivateKey() {
...
+    useState(false);
+
+  const checkAccount = useCallback(
```
meeh0w, line 61:
```suggestion
  const checkIfAccountExists = useCallback(
```

src/components/ledger/LedgerConnector.tsx
```diff
@@ -110,12 +116,42 @@ export function LedgerConnector({
...
+          pubKeys: publicKeys,
+          secretType:
+            derivationPath === DerivationPath.BIP44
+              ? SecretType.Ledger
+              : SecretType.LedgerLive,
+          dryRun: true,
+        });
+      } catch (e) {
+        setIsLedgerExistsError(true);
+        throw new Error('wallet already exists');
```
meeh0w, line 135: We should check the error message - there are other reasons `importLedger(...)` could fail.

src/components/ledger/LedgerConnector.tsx
```diff
@@ -48,11 +50,13 @@ export interface LedgerConnectorData {
 interface LedgerConnectorProps {
   onSuccess: (data: LedgerConnectorData) => void;
   onTroubleshoot: () => void;
+  isWalletExist?: boolean;
 }
 
 export function LedgerConnector({
   onSuccess,
   onTroubleshoot,
+  isWalletExist,
```
meeh0w, line 59: If I understand correctly, this prop tells `LedgerConnector` to verify if the secret already exists, right?

If so, I think we should rename it. Right now it sounds more as a statement than a question/request to check for something.

Maybe: `checkIfWalletExists` or `withWalletDeduplicationWarning` ?

meeh0w (changes requested):

meeh0w (approved):

bferenc (approved):

meeh0w:
(I also tested, btw.)

bferenc (approved):

meeh0w (approved):

PR #93, opened 2024-11-19
feaT: CP-9487 CP-9473 ledger flow

vvava:
## Description
We want to polish a bit our ledger wallet flow:
1. We want to show the addresses with a balance or the first 3 addresses
2. If the user has no ledger device detected, give the user the opportunity to close the page and reopen the extension

## Changes

Address list displaying and the extension reopen.

## Testing

Go to account page -> try to add a ledger wallet with a ledger connected and without that.

## Screenshots:

https://github.com/user-attachments/assets/510721e6-58da-4d9d-828b-ba2cb1b4f0d3

https://github.com/user-attachments/assets/cb34bfce-d619-4af5-8cef-43ba4cb094f1

## Checklist for the author


bferenc (changes requested):
Thanks, left some comments and have 2 additional questions:
- is this based on https://github.com/ava-labs/core-extension/pull/90? If so, can you please update this branch and compare it against that one?
- so we simply hide the 0 balance addresses during onboarding?
  - what if none of the first 3 accounts have balances?
  - shouldn't we stop when an account has 0 balance? (e.g.: don't show 3rd account if the 2nd doesn't have AVAX)
  - is it expected to still onboard with all 3 public keys for LEDGER_LIVE regardless of their balances?

vvava:
> Thanks, left some comments and have 2 additional questions:
> 
> * is this based on [feat: CP-9472 add duplicated account dialog #90](https://github.com/ava-labs/core-extension/pull/90)? If so, can you please update this branch and compare it against that one?
> * so we simply hide the 0 balance addresses during onboarding?
>   
>   * what if none of the first 3 accounts have balances?
>   * shouldn't we stop when an account has 0 balance? (e.g.: don't show 3rd account if the 2nd doesn't have AVAX)
>   * is it expected to still onboard with all 3 public keys for LEDGER_LIVE regardless of their balances?

It is based on that ticket yes, I will revisit if I will able to merge in to the main.

```
AC1: when the user is connecting a Ledger only show the accounts with a C-Chain balance

if that is 1 account show one, if there are 3 accounts show 3

cap the forward look at 3 accounts 
```

As I understand the ticket we should display the addresses with any balance to the user up to three addresses (so if the 1st and the 3rd accounts have some AVAX we should hide only the 2nd one but not to stop getting the 3rd address). If we cannot find any balances we should show all the three addresses as we are doing at the moment.  
I don't think we need to do any other changes about the current behaviour of the onboarding / adding a new ledger wallet.

src/background/runtime/BackgroundRuntime.ts
```diff
@@ -13,7 +12,6 @@ export class BackgroundRuntime {
     private connectionService: ConnectionService,
     private lockService: LockService,
     private onboardingService: OnboardingService,
-    private bridgeService: BridgeService,
```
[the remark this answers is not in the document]
vvava: I'm not sure, it was not used anywhere so it seemed to be just hanging there as an unnecessary dependency.

bferenc (commented):
Can you fix the merge conflict, please?

bferenc (dismissed):

src/background/runtime/BackgroundRuntime.ts
```diff
@@ -4,15 +4,16 @@ import { ConnectionService } from '@src/background/connections/ConnectionService
 import { singleton } from 'tsyringe';
 import { LockService } from '@src/background/services/lock/LockService';
 import { OnboardingService } from '@src/background/services/onboarding/OnboardingService';
-import { BridgeService } from '@src/background/services/bridge/BridgeService';
 import { ModuleManager } from '../vmModules/ModuleManager';
+import { BridgeService } from '@avalabs/bridge-unified';
```
meeh0w, line 8:
```suggestion
import { BridgeService } from '../services/bridge/BridgeService';
```

src/pages/Accounts/AddWalletWithLedger.tsx
```diff
@@ -306,15 +308,38 @@ export function AddWalletWithLedger() {
...
+                  <Stack sx={{ width: '50%' }}>
+                    <Tooltip
+                      title={t(
+                        'Clicking the cancel button will close the tab and open the extension for you. If the extension doesn’t open automatically, please open it manually.'
+                      )}
+                    >
+                      <Button
+                        color="secondary"
+                        onClick={() => {
+                          browser.action.openPopup();
```
meeh0w, line 325: This is a great addition! Let's also do it on the success screen, when users click "Close" 🎉

meeh0w (changes requested):

meeh0w (approved):

bferenc (approved):

PR #98, opened 2024-11-25
fix: price warning always visible

meeh0w:
## Description

For the missing price warning, we look at all the tokens, even those with `balance: 0n`. We also take into consideration the tokens that the user has hidden.

This provides a way to fix the price warning, which right now is visible basically all the time right now.

## Testing
* If you get the missing price warning, hide the spam/testing tokens or tokens which don't have a price.

## Screenshots:

https://github.com/user-attachments/assets/45b77ab5-12f1-45cf-8c39-e6299515f766

## Checklist for the author


gergelylovas (approved):

bferenc (approved):
🎉

PR #94, opened 2024-11-20
feat: new account switcher

meeh0w:
## Description
* [CP-9490](https://ava-labs.atlassian.net/browse/CP-9490)

## Changes
* New account switcher designs
* New account details view
* ~_No balance rollup yet (AC3), will be done in a separate PR_~
    * Done in #101 

## Testing
* Play with the account switcher, try to break it

## Screenshots:

https://github.com/user-attachments/assets/b639e4d8-3e80-4b83-b8ac-b683faebd8aa

## Checklist for the author
- [ ] I've covered new/modified business logic with Jest test cases.
- [x] I've tested the changes myself before sending it to code review and QA.

PR #101, opened 2024-11-27
feat: wallet balance rollup

meeh0w:
## Description
* [CP-9513](https://ava-labs.atlassian.net/browse/CP-9513)
* ⚠️ builds on top of #94 

## Changes
* Total balance for the wallet is added in the account switcher

## Testing
* Have underived accounts with balances (i.e. perform a clean onboarding)
* In the account switcher, next to the wallet name, you should see the total balance for the wallet. It should be a total of:
    * Balances for all derived accounts on all favorited networks (C/P/X chains always included)
    * X/P-Chain balances for underived accounts (i.e. if you have only two accounts derived, but the 3rd also has some balance on the X- or P-Chain, this balance be included in the total).
        * _**Yes**, this means that summing up the balances of all visible accounts for the wallet may not always result in the same value as the balance next to the wallet name_
* If there are accounts with X/P-Chain balances that are not visible in the account switcher (not derived yet), `View P-Chain Details` link should be displayed

## Screenshots:

https://github.com/user-attachments/assets/446be001-ba4c-4e3d-8731-b403810580b9

## Checklist for the author


src/contexts/BalancesProvider.tsx
```diff
@@ -236,11 +234,11 @@ export function BalancesProvider({ children }: { children: any }) {
       return;
     }
 
-    if (isPolling) {
-      const tokenTypes = Object.entries(subscribers)
-        .filter(([, subscriberCount]) => subscriberCount > 0)
-        .map(([tokenType]) => tokenType as TokenType);
+    const tokenTypes = Object.entries(subscribers)
```
meeh0w, line 237: This fixes a bug which was causing us to still actively poll balances on screens which don't necessarily need them.

src/background/services/balances/BalanceAggregatorService.ts
```diff
@@ -143,14 +144,15 @@ export class BalanceAggregatorService implements OnLock, OnUnlock {
       for (const [chainId, chainBalances] of freshData) {
         for (const [address, addressBalance] of Object.entries(chainBalances)) {
           aggregatedBalances[chainId] = {
+            ...aggregatedBalances[chainId],
```
meeh0w, line 147: This fixes a bug which was causing existing balances for the address to not be kept in cache when updating with partial values.

src/pages/Home/components/Portfolio/NetworkWidget/PchainActiveNetworkWidgetContent.tsx
meeh0w, line 1: Changes in this file fix a bug in the active network widget for P-Chain (we stopped showing the balance breakdown):
image

src/background/services/balances/utils/calculateTotalBalance.ts
meeh0w, line 1: This was no longer used

src/background/services/balances/handlers/getTotalBalanceForWallet/helpers/getIncludedNetworks.ts
```diff
@@ -0,0 +1,28 @@
+import { uniq } from 'lodash';
+import { ChainId } from '@avalabs/core-chains-sdk';
+
+import { ChainListWithCaipIds } from '@src/background/services/network/models';
+
+export function getIncludedNetworks(
```
gergelylovas, line 6: https://github.com/user-attachments/assets/7ca76ead-8ec9-405a-bf81-9c619790a792

Something is off when the user is on P-Chain and is switching between P and C. I have P chain favorited

src/background/services/balances/handlers/getTotalBalanceForWallet/getTotalBalanceForWallet.ts
```diff
@@ -0,0 +1,160 @@
...
+      let totalBalanceInCurrency = calculateTotalBalanceForAccounts(
+        derivedAddressesBalances,
+        derivedAccounts,
+        networksIncludedInTotal
+      );
+      let hasBalanceOnUnderivedAccounts = false;
+
+      if (underivedAccounts.length > 0) {
+        // Get balance for underived addresses for X- and P-Chain.
+        // We DO NOT cache this response. When fetching balances for multiple X/P addresses at once,
```
gergelylovas, line 126: Should we add a follow up to cache this response? Seems like it can take quite long for people with many UTXOs

gergelylovas (changes requested):

src/background/services/balances/handlers/getTotalBalanceForWallet/helpers/getIncludedNetworks.ts
meeh0w: I think it's because C-Chain is not _actually_ a favorite network - we just always show it. Let me try to fix that.

src/background/services/balances/BalanceAggregatorService.ts
meeh0w: I'll take another look at it 👀

meeh0w: Yeah, we needed something like this instead: 24f970b16

src/pages/Accounts/components/WalletHeader.tsx
```diff
@@ -109,7 +117,11 @@ export default function WalletHeader({
           textAlign="end"
           color="text.secondary"
         >
-          {/* TODO: total balance of the entire wallet */}
+          {isLoading ? (
```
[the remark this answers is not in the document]
meeh0w: This fixes it: 171a8e6fb

src/background/services/balances/handlers/getTotalBalanceForWallet/helpers/getAccountsWithActivity.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      })
+      // Second batch for internal addresses
+      .mockResolvedValue({
+        gap: 21,
+        result: [],
+      });
+    // mockedActivityFetcher
+    //   .mockImplementationOnce(makeActiveAddresses) // First batch, external addresses
+    //   .mockImplementationOnce(makeActiveAddresses) // First batch, internal addresses
+    //   .mockImplementation(() => ({ addresses: [] })); // All subsequent batches
```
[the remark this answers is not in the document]
meeh0w: f0e102fb0

src/background/services/balances/handlers/getTotalBalanceForWallet/helpers/getIncludedNetworks.ts
meeh0w: Yeah, so that was actually the case (fixed in 634ca1643).

Since C-Chain is not really a favorite network, it's not included in the total balance if it's not the active network. See the prod build on Bitcoin, for example:

image

versus with the fix (C-Chain balance included):

image

src/background/services/balances/handlers/getTotalBalanceForWallet/getTotalBalanceForWallet.ts
meeh0w: Ticket: https://ava-labs.atlassian.net/browse/CP-9628

gergelylovas (approved):
Nice work!

PR #104, opened 2024-12-06
fix: provide default value to convert

meeh0w:
## Description
_N/A_

## Changes
* We were calling `stringToBigint(...)` on an empty string

## Testing
* Go to Swap and enter an amount, it should not crash

## Checklist for the author


gergelylovas (approved):

bferenc (approved):
issue gone ✅

PR #110, opened 2024-12-12
feat: bring back address copy button

meeh0w:
* [CP-9683](https://ava-labs.atlassian.net/browse/CP-9683)

## Description
* Brings back the copy address button to the account switcher.

## Testing
* Use the address copy button and make sure it works
* Bonus points for verifying the captured event names

## Checklist for the author

### DIFF
--- a/src/pages/Accounts/components/AccountItem.tsx
+++ b/src/pages/Accounts/components/AccountItem.tsx
@@ -1,6 +1,9 @@
 import {
   Checkbox,
   Collapse,
+  CopyIcon,
+  Grow,
+  IconButton,
   Stack,
   Tooltip,
   Typography,
@@ -15,6 +18,7 @@ import {
   useState,
 } from 'react';
 import { useTranslation } from 'react-i18next';
+import { NetworkVMType } from '@avalabs/core-chains-sdk';
 
 import { useBalancesContext } from '@src/contexts/BalancesProvider';
 import { useAccountsContext } from '@src/contexts/AccountsProvider';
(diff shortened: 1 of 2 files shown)


src/pages/Accounts/components/AccountItem.tsx
```diff
@@ -299,3 +334,18 @@ export const AccountItem = forwardRef(
...
+      return 'AccountSelectorBtcAddressCopied';
+    case NetworkVMType.EVM:
+      return 'AccountSelectorEthAddressCopied';
+    case NetworkVMType.AVM:
+      return 'AccountSelectorAVMAddressCopied';
+    case NetworkVMType.PVM:
+      return 'AccountSelectorPVMAddressCopied';
+    default:
+      return 'AccountSelectorAddressCopied';
+  }
```
gergelylovas, line 350: Are these older event names we need to follow? Can we do something like this instead?
```suggestion
return  `AccountSelector${type}AddressCopied`;
```

gergelylovas (approved):
The copy button works well. I wonder if we could make the address selectable in the tooltip as well. I think it was possible in one of the previous iterations of the current redesign. 

https://github.com/user-attachments/assets/3efc41e4-7663-4365-8eb9-b8dda394caa9

src/pages/Accounts/components/AccountItem.tsx
meeh0w: We used two of them before: `Btc` and `Eth`

meeh0w: Simplified a bit in 6b13e6aae + made the tooltip contents selectable.

gergelylovas (approved):

vvava (approved):

PR #111, opened 2024-12-13
Fix: CP-9487 && CP-9637 accounts and wallets changes

vvava:
## Description
Contains: 
[CP-9637](https://ava-labs.atlassian.net/jira/software/c/projects/CP/boards/72?selectedIssue=CP-9637)
[CP-9487](https://ava-labs.atlassian.net/jira/software/c/projects/CP/boards/72?selectedIssue=CP-9487)

## Changes

When the ledger wallet added we now add only the accounts with balance to the extension.
After import a wallet via seed phrase, ledger, keystore we now make the first account of that wallet to be the activa account.

## Testing

Create and import wallets with ledger and mnemonic.

## Screenshots:

Ledger BIP44:

https://github.com/user-attachments/assets/f5c7555e-52eb-4549-9157-eed89947bab4

Ledger LedgerLive:

https://github.com/user-attachments/assets/6b8c8af6-1338-40ec-b003-4b5df82650f4

Seed phrase import:

https://github.com/user-attachments/assets/ce78871b-873c-456b-b200-d10a2591f260

## Checklist for the author


src/pages/Accounts/hooks/usePrivateKeyImport.ts
```diff
@@ -21,6 +21,7 @@ export const usePrivateKeyImport = () => {
           importType: ImportType.PRIVATE_KEY,
           data: utils.strip0x(privateKey),
         });
+        await selectAccount(accountId);
```
meeh0w, line 24: We should have it outside of the `importPrivateKey()` function, because it can be called multiple times in loops, for example by `useKeystoreFileImport()`:

https://github.com/ava-labs/core-extension/blob/85cbf294315230af73e94e77b8cc439049844de0/src/pages/Accounts/hooks/useKeystoreFileImport.ts#L48-L65


This will activate all accounts one after another, which will cause lots of re-renders and we'll start polling balances for all of them at the same time, which will be a lot of requests.

src/pages/Accounts/AddWalletWithLedger.tsx
```diff
@@ -102,6 +104,7 @@ export function AddWalletWithLedger() {
 
   const handleImport = useCallback(
     async (name?: string) => {
+      console.log('handleImport: ', handleImport);
```
meeh0w, line 107: ✂️ 

```suggestion
```

src/components/ledger/LedgerConnector.tsx
```diff
@@ -81,6 +82,8 @@ export function LedgerConnector({
   const [addresses, setAddresses] = useState<AddressType[]>([]);
   const [hasPublicKeys, setHasPublicKeys] = useState(false);
   const [dropdownDisabled, setDropdownDisabled] = useState(true);
+  const lastAccountIndexWithBalance = useRef(0);
```
meeh0w, line 85: Why `useRef()` instead of `useState()`?

src/components/ledger/LedgerConnector.tsx
```diff
@@ -98,7 +102,12 @@ export function LedgerConnector({
       addressList: AddressType[] = []
     ) => {
       const address = getAddressFromXPub(xpubValue, accountIndex);
+
       const { balance } = await getAvaxBalance(address);
+
+      if (balance.balanceDisplayValue && balance.balanceDisplayValue !== '0') {
+        lastAccountIndexWithBalance.current = accountIndex;
+      }
```
meeh0w, line 110: This only works if `getAddressFromXpubKey()` is always called in the correct order (which can be easily broken and would result in a bug hard to fix).

We should check if `accountIndex` truly is greater than `lastAccountIndexWithBalance.current`.

**EDIT:**
I just noticed we add `balance` in `setAddresses(...)` call just below. Would it not be easier to get the `lastAccountIndexWithBalance` by doing something like this:

```ts
// have to do Math.max(), because findLastIndex() may return -1
const lastIndexWithBalance = Math.max( 
  0,
  addressList.findLastIndex((addr) => addr.balance !== '0'),
);
```

❓

meeh0w (changes requested):
I left some ideas 👍

src/components/ledger/LedgerConnector.tsx
vvava: We don't display that value just use it in calculation therefor rerendering is not necessary. It saves the cost of rerendering and other side effects of that.

src/components/ledger/LedgerConnector.tsx
```diff
@@ -81,6 +82,9 @@ export function LedgerConnector({
   const [addresses, setAddresses] = useState<AddressType[]>([]);
   const [hasPublicKeys, setHasPublicKeys] = useState(false);
   const [dropdownDisabled, setDropdownDisabled] = useState(true);
+  // TODO: with lastindex
```
gergelylovas, line 85: Please remove the TODO comment

src/background/services/onboarding/handlers/ledgerOnboardingHandler.ts
```diff
@@ -46,8 +47,16 @@ export class LedgerOnboardingHandler implements HandlerType {
...
+    const {
+      xpub,
+      xpubXP,
+      pubKeys,
+      password,
+      analyticsConsent,
+      walletName,
+      numberOfAccountsToCreate,
+    } = (request.params ?? [])[0] ?? {};
+    console.log('numberOfAccountsToCreate: ', numberOfAccountsToCreate);
```
gergelylovas, line 59: Please remove

src/background/services/onboarding/handlers/ledgerOnboardingHandler.ts
```diff
@@ -92,20 +101,12 @@ export class LedgerOnboardingHandler implements HandlerType {
       };
     }
 
-    if (xpub) {
+    for (let i = 0; i < (numberOfAccountsToCreate || 1); i++) {
```
gergelylovas, line 104: You should check for `pubKeys.length` as well otherwise it might attempt to add an account we don't have the public key for

src/background/services/wallet/handlers/importLedger.ts
```diff
@@ -105,10 +117,7 @@ export class ImportLedgerHandler implements HandlerType {
       });
     }
 
-    const numberOfAccountsToCreate =
-      secretType === SecretType.LedgerLive ? pubKeys?.length : undefined;
-
-    await this.#addAccounts(id, numberOfAccountsToCreate);
+    await this.#addAccounts(id, numberOfAccountsToCreate || 3);
```
gergelylovas, line 120: Please add a check to make sure that the `numberOfAccountsToCreate` is <= than `pubKeys?.length` when we don't have the xPub.

gergelylovas (changes requested):

src/background/services/wallet/handlers/importLedger.ts
```diff
@@ -117,7 +117,14 @@ export class ImportLedgerHandler implements HandlerType {
       });
     }
 
-    await this.#addAccounts(id, numberOfAccountsToCreate || 3);
+    const defaultAccountsToCreate = 3;
+    const accountsToCreate = !pubKeys
+      ? numberOfAccountsToCreate || defaultAccountsToCreate
+      : numberOfAccountsToCreate || defaultAccountsToCreate <= pubKeys.length
+        ? numberOfAccountsToCreate || defaultAccountsToCreate
+        : pubKeys.length;
```
meeh0w, line 125: So, this repeats `numberOfAccountsToCreate || defaultAccountsToCreate` three times, which makes it a bit hard to read. Can we simplify?

If we have `pubKeys` (Ledger Live), we want to create `<numberOfAccountsToCreate>` accounts, but no more than `pubKeys.length` and never more than 3, right?

And if we don't have `pubKeys` (BIP44), we want to create `<numberOfAccountsToCreate> (if known and above zero), `<defaultAccountsToCreate>` otherwise, but still never more than 3?

So how about this?

```suggestion
    const accountsToCreate = Math.min(
      3, // never create more than 3 accounts
      pubKeys // if we use Ledger Live...
        ? Math.min(pubKeys.length, accountsToBeCreated) // ...go with accountsToBeCreated, but never exceed pubKeys.length...
        : accountsToBeCreated // ... and if we're on BIP44, go with accountsToBeCreated
    );
```

meeh0w (approved):

PR #112, opened 2024-12-13
Yarn version upgrade

csabbee:
## Description

The aim of this PR is to update the version of the yarn package manager to the latest stable version: 4.5.3

## Changes

* setting `packageManager` property in the `package.json` file
* modifying entries under `resolutions` (This was required because the previous setup got removed when running `yarn install` after the yarn version upgrade)
  *  `minimatch`
  *  `nth-check`
  *  `glob-parent`
* entries added under `allowScripts` which was visible in the console after running the `postinstall` script

## Testing

I run all the scripts in `package.json`, apart from the `sentry` (which I assume needs some additional setup) run without any problem.

Obviously some additional testing is required, to really make sure nothing is broken

## Checklist for the author

- [ ] ~~I've covered new/modified business logic with Jest test cases.~~ N/A


gergelylovas (changes requested):
Thank you for the updates!
Seems like some changes are needed to the `.github/workflow` files to pass CI checks. `Error: Cannot find module '/home/runner/work/core-extension/core-extension/.yarn/releases/yarn-4.5.3.cjs'`.

PR #115, opened 2024-12-18
feat: (dApp Handler) CP-9471 expose delete accounts endpoint 

frichards:
## Description
Ticket: https://ava-labs.atlassian.net/browse/CP-9471
Figma: https://www.figma.com/design/KgVJGkC7vEMh9CZtidijTW/Extension?node-id=24020-246539&t=uRFV5H51uyanBo7j-0

As a new feature, core web needs to be able to request the accounts to be removed from extension. 

## Changes
1. Adding avalanche_deleteAccounts handler
2. Adding DeleteAccount approval screen

Minor change
The font weight for ApprovalSectionHeader has been requested to be changed to 600.

## Testing
From Core web: 
1. Request avalanche_deleteAccounts.
2. The accounts should be removed without any approval screen

From Non-Core web:
1. Request avalanche_deleteAccounts
2. The approval screen should pop up

Just like extension, for primary accounts in a wallet, the account with largest index can be deleted. If not the largest, it will throws an error. 
Also If the request ends up deleting every accounts, it also throws an error.

## Screenshots:
Screenshot 2024-12-18 at 13 27 36

## Checklist for the author

- [X] I've covered new/modified business logic with Jest test cases.


src/background/services/accounts/handlers/avalanche_deleteAccounts.ts
```diff
@@ -0,0 +1,235 @@
...
+  imported: ImportedAccount[];
+  wallet: {
+    [walletId: string]: string;
+  };
+};
+
+type Params = [accountIds: string[]];
+
+@injectable()
+export class AvalancheDeleteAccountsHandler extends DAppRequestHandler<
```
meeh0w, line 34: Can you also remove the [`accounts/handlers/deleteAccounts.ts`](https://github.com/ava-labs/core-extension/blob/17924c76cc62eb23d9a6b6e8ed70446b89c8ea86/src/background/services/accounts/handlers/deleteAccounts.ts#L13) handler and make `AccountsProvider` use your new handler instead? 🙂

We already got rid of the old `renameAccount.ts` when we were adding the [`AvalancheRenameAccountHandler.ts`](https://github.com/ava-labs/core-extension/blob/85cbf294315230af73e94e77b8cc439049844de0/src/background/services/accounts/handlers/avalanche_renameAccount.ts#L20) so Core Web could use it too and we now have just one handler which Extension's UI also uses:

https://github.com/ava-labs/core-extension/blob/85cbf294315230af73e94e77b8cc439049844de0/src/contexts/AccountsProvider.tsx#L107-L115


We could benefit from the logic & guards you added here (i.e. the old handler does not check if only the last accounts are being removed 🫣).

src/pages/Wallet/DeleteAccounts.tsx
```diff
@@ -0,0 +1,202 @@
...
+            {t(
+              'Removing the account will delete all local account information stored on this computer. Your assets on chain will remain on chain.'
+            )}
+          </Typography>
+        </Stack>
+        <Stack direction="row" sx={{ gap: 1 }}>
+          <Button
+            color="secondary"
+            size="large"
+            data-testid="switch-account-reject-btn"
```
meeh0w, line 174:
```suggestion
            data-testid="delete-accounts-reject-btn"
```

src/pages/Wallet/DeleteAccounts.tsx
```diff
@@ -0,0 +1,202 @@
...
+            onClick={() => {
+              cancelHandler();
+              window.close();
+            }}
+            fullWidth
+          >
+            {t('Reject')}
+          </Button>
+          <Button
+            data-testid="switch-account-approve-btn"
```
meeh0w, line 184:
```suggestion
            data-testid="delete-accounts-approve-btn"
```

meeh0w (changes requested):
Great work! I added some minor comments, nothing too serious 👏

meeh0w (approved):
Nicely done 👏

src/background/connections/dAppConnection/models.ts
```diff
@@ -34,6 +34,7 @@ export enum DAppProviderRequest {
   AVALANCHE_SET_DEVELOPER_MODE = 'avalanche_setDeveloperMode',
   ACCOUNT_SELECT = 'avalanche_selectAccount',
   ACCOUNT_RENAME = 'avalanche_renameAccount',
+  ACCOUNTS_DELETE = 'avalanche_deleteAccounts',
```
gergelylovas, line 37: Should we start prefixing these methods with `wallet_` instead of `avalanche_`?
My reasoning is that this and some other methods are wallet related while `avalanche_` implies network specificity like `eth_` does

gergelylovas (approved):

src/background/connections/dAppConnection/models.ts
frichards: Oh I see! Yes. I will update it to be wallet_deleteAccounts 😄 
The other ones we should change to is selectAccount and renameAccount?

frichards: Talked to Michal and Gergely, the name change from avalanche_ to wallet_ is going to happen later on a separate ticket

PR #128, opened 2025-01-29
feat: CP-9655 check the fraction numebr and transform

vvava:
## Description

The fraction number display was not the AC.

## Changes

If the fraction has zeros the first non-zero value will be displayed (when the integer is 0, otherwise we will show two digits).

## Testing

Go the the portfolio and a transaction (e.g. send) and see the value has the right fraction numbers.

## Screenshots:

image

## Checklist for the author


meeh0w (changes requested):
Let's add some unit tests for the formatter so we don't resurrect these issues in the future 🙏

meeh0w (approved):

bferenc (approved):

PR #132, opened 2025-02-03
fix: token visibility

meeh0w:
## Description
* [CP-9861](https://ava-labs.atlassian.net/browse/CP-9861)

## Changes
* Stops ignoring the token data returned by `TokenManager` :)
* Filters hidden tokens from token lists on Send/Swap/Bridge (overridable)
* Fixes the EVM signer for Unified Bridge (it was ignoring `value` prop on the transaction payload)

## Testing
* Add a custom token, e.g. (`0x13b1f0579bc895b2ffb835f295fd9b63fef36da0`) (you should succeed)
* Hide a token (with balance) on the `Manage Tokens` list and then make sure it's not visible on Send/Swap/Bridge.

## Checklist for the author

### DIFF
--- a/src/contexts/UnifiedBridgeProvider.tsx
+++ b/src/contexts/UnifiedBridgeProvider.tsx
@@ -167,7 +167,7 @@ export function UnifiedBridgeProvider({
   const evmSigner: EvmSigner = useMemo(
     () => ({
       sign: async (
-        { from, data, to },
+        { from, data, to, value },
         _,
         { currentSignature, requiredSignatures },
       ) => {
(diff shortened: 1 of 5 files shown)


src/contexts/UnifiedBridgeProvider.tsx
```diff
@@ -624,19 +638,24 @@ export function UnifiedBridgeProvider({
...
+          gasSettings,
+        });
 
-      await trackBridgeTransfer(bridgeTransfer);
+        await trackBridgeTransfer(bridgeTransfer);
 
-      return bridgeTransfer.sourceTxHash;
+        return bridgeTransfer.sourceTxHash;
+      } catch (err) {
+        console.log('DEBUG', err);
```
bferenc, line 656:
```suggestion
        console.error(err);
```

bferenc (approved):
tested as well ✅ 
just one nit below 👇

bferenc (approved):

vvava (approved):

PR #133, opened 2025-02-04
refactor: change how addresses are derived

meeh0w:
## Description
* [CP-9730](https://ava-labs.atlassian.net/browse/CP-9730)
* Related VM Modules PR: https://github.com/ava-labs/vm-modules/pull/223

## Changes
_description incoming_

## Testing
_description incoming_

## Screenshots:
_No visual change_

## Checklist for the author


src/tests/test-utils.tsx
```diff
@@ -38,6 +40,21 @@ export const buildRpcCall = <M extends string>(
     request: payload,
   }) as const;
 
+export const expectToThrowErroCode = async (
```
meeh0w, line 43: typo

src/background/vmModules/ApprovalController.ts
```diff
@@ -70,6 +78,19 @@ export class ApprovalController implements BatchApprovalController {
     // Transaction Reverted. Show a toast? Trigger browser notification?',
   };
 
+  async requestPublicKey({
+    curve,
+    derivationPath,
+    secretId,
+  }: RequestPublicKeyParams): Promise<string> {
+    // TODO: Ask user approval if needed?
```
meeh0w, line 86: ✂️ approval not needed at this point

src/background/services/storage/schemaMigrations/migrations/wallet_v5/newSchema.ts
meeh0w, line 1: ⚠️ Crucial unit tests are missing here.

src/background/services/storage/schemaMigrations/migrations/wallet_v5/wallet_v5.ts
meeh0w, line 1: - [ ] ⚠️ Crucial unit tests are missing here.

src/background/services/secrets/models.ts
```diff
@@ -26,6 +27,28 @@ export enum SecretType {
   Fireblocks = 'fireblocks',
 }
 
+export type Secp256k1 = 'secp256k1';
+export type Ed25519 = 'ed25519';
+export type Curve = Secp256k1 | Ed25519;
+export const EVM_BASE_DERIVATION_PATH = "m/44'/60'/0'";
+export const AVALANCHE_BASE_DERIVATION_PATH = "m/44'/9000'/0'";
```
meeh0w, line 34: These could probably use a better name (those are paths for the legacy `xpub` and `xpubXP`).

src/background/services/secrets/AddressPublicKey.ts
```diff
@@ -0,0 +1,213 @@
...
+    switch (curve) {
+      case 'secp256k1': {
+        const seedNode = fromSeed(seed);
+        key = hex.encode(seedNode.derivePath(derivationPath).publicKey);
+        break;
+      }
+
+      case 'ed25519': {
+        const hdKey = slip10.fromMasterSeed(seed);
+        key = hex.encode(hdKey.derive(derivationPath).publicKey);
```
meeh0w, line 198: Use `.publicKeyRaw` here -- `publicKey` is 0-prefixed (33 bytes total), we need the raw 32 bytes.

csabbee (approved):
Kudos for this, really well done!
I'm not saying I did an all around testing but, I did played around with the dev build and didn't run into anything 🔥

src/background/services/storage/schemaMigrations/migrations/wallet_v5/wallet_v5.test.ts
```diff
@@ -0,0 +1,379 @@
...
+    xpubXP: 'xpubXP',
+  };
+  const mockLedgerSecrets: Legacy.LedgerSecrets = {
+    id: 'ledgy',
+    derivationPath: 'bip44',
+    name: 'Ledger Wallet',
+    xpub: 'xpub',
+    xpubXP: 'xpubXP',
+    secretType: 'ledger',
+    btcWalletPolicyDetails: {
```
gergelylovas, line 27: Should we have test scenarios where the optional properties are missing?

src/background/services/storage/schemaMigrations/migrations/wallet_v5/wallet_v5.ts
```diff
@@ -0,0 +1,323 @@
...
+  return {
+    importedAccounts: currentSecrets.importedAccounts,
+    wallets: primaryWallets,
+    version: VERSION,
+  };
+};
+
+export default {
+  previousSchema: legacySecretsSchema,
+  dependencyKeys: [ACCOUNTS_STORAGE_KEY],
```
gergelylovas, line 317: How do we ensure version consistency here?
Since the ACCOUNTS_STORAGE_KEY's migrations are running earlier than the wallet migrations (for now) it is possible that a future migration on the account storage will break this V5 migration. We should make sure the key's migration order stays the same.
Maybe we should use the types from the account service? Then at least a change in the types would trigger a TS error in here.

gergelylovas (changes requested):
Very nice work. Especially like the documentation and the new clean secret storage model.

src/background/services/storage/schemaMigrations/migrations/wallet_v5/wallet_v5.ts
meeh0w: Very good point. I'm not sure how to approach this...

We could create a method for `StorageService` that would allow loading a specific version of a given storage key, but that would only work if the migrations had both `up` and `down` methods (so if `Accounts` are on version `4`, we can still map it to version `3`).

So how about this:

1. We use `Accounts` model here as you suggested
2. We leave a comment in this `wallet_v5` migration that says:

> If this throws type errors due to `Accounts` model being updated, you're likely introducing a migration for `ACCOUNTS_STORAGE_KEY`. There are a few things that need to be done:
> 
> 1. The new migration needs to have both `up` and `down` methods, so that once `Accounts` is migrated to v4+, other places (like this migration here) can still request v3.
> 2. The `StorageService` needs a way to load a key with requested version (i.e. `StorageService.loadVersion(storageKey: string, version: number)`). If the key is at specified version already, just return it. If it's already been migrated to a higher version - use `migration.down()` to go back (don't store it, just return it).
> 3. This migration (`wallet_v5`) needs to be re-written so that it uses the new `StorageService.loadVersion()` method.

src/background/vmModules/ModuleManager.ts
```diff
@@ -51,6 +51,10 @@ export class ModuleManager {
     this.#_modules = modules;
   }
 
+  get modules(): Module[] {
+    return this.#modules;
+  }
+
```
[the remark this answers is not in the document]
meeh0w: Hmm 🤔 I managed to drop it entirely, I'm now loading the modules the same way for address derivation and building derivation paths: b8a9c5429

src/background/services/storage/schemaMigrations/migrations/wallet_v5/wallet_v5.test.ts
meeh0w: Great suggestion, caught a bug :) 79c17a149

Thanks!

src/background/vmModules/ModuleManager.ts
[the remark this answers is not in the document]
meeh0w: Aaand I brought it back: f7351e6f2.

It's been some time and I completely forgot the reason I need to iterate over all registered modules is because I want to build derivation paths for as many VM types as possible whenever user adds an account.

If I build derivation paths only for the active networks, that means that we never derive pub keys for HVM, for example. Unless we want to make that derivation on-demand (i.e. when user switches to a network that needs it) which makes sense, this needs to stay imo. And even then, I wouldn't want to do it within this PR.

LMK what you think :)

src/background/services/secrets/AddressResolver.ts
```diff
@@ -0,0 +1,128 @@
...
+  #moduleManager?: ModuleManager;
+
+  constructor(
+    private readonly networkService: NetworkService,
+    private readonly secretsService: SecretsService,
+  ) {}
+
+  init(moduleManager: ModuleManager) {
+    this.#moduleManager = moduleManager;
+  }
```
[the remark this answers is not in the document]
gergelylovas: Okay, let's not add more refactoring to this PR but create a separate ticket for this. 
Can you please leave a comment on the code just so we know?

src/background/services/storage/schemaMigrations/migrations/wallet_v5/wallet_v5.ts
gergelylovas: Sounds good, let's do this for now.

src/background/vmModules/ModuleManager.ts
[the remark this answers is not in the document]
gergelylovas: okay let's leave it as is.

gergelylovas (approved):

csabbee (approved):
Tested by setting up a bunch of accounts with names, also with physical ledger connected.
After installing the extension from this branch, everything was still there, so the migration worked.
Tested sending, swapping, signing, everything is working.

src/background/services/secrets/AddressResolver.ts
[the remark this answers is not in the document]
meeh0w: Added a comment: 653f69163 and created a ticket: [CP-10132](https://ava-labs.atlassian.net/browse/CP-10132)

csabbee (approved):

gergelylovas (approved):

PR #136, opened 2025-02-05
chore: CP-9872: modify eslint config

csabbee:
## Description

The aim of this PR is to introduce the [consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports/) config to eslint

JIRA ticket: [CP-9872](https://ava-labs.atlassian.net/browse/CP-9872)

This helps webpack with the built because it can "throw" out type imports so there is no remnant of it at runtime (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html)

## Changes

* Adding the rule to the `eslint.config.mjs` file
* fixing the type imports in the files
* modifying the types in `src/background/services/actions/models.ts` because the type check was failing

### UPDATE
* fixing test in `src/background/services/accounts/handlers/avalanche_selectAccount.test.ts`
  * after the type change this test started failing, saying that the DI container couldn't locate any instance of the `ApprovalService`.

### UPDATE #2 
Because of the DI, there were runtime problems because of the type import conversion
Fortunately there is a configuration change for it:
https://typescript-eslint.io/blog/changes-to-consistent-type-imports-with-decorators
So this rule will skip all the files that have any type of `@decorators` in them. Because of this exclusion, there is "only" 613 file changes instead of the 800+ 😅 

## Testing

Since the changes introduced here are only "cosmetics", namely only changing the eslint rules, the testing was done by running `linting` and `typechecking`

## Checklist for the author

- [] ~I've covered new/modified business logic with Jest test cases.~


src/background/services/accounts/handlers/avalanche_selectAccount.test.ts
```diff
@@ -27,6 +28,11 @@ describe('background/services/accounts/handlers/avalanche_selectAccount.ts', ()
   } as any;
 
   container.registerInstance(ActionsService, actionsServiceMock as any);
+  const actionService = container.resolve(ActionsService);
+  container.registerInstance(
+    ApprovalService,
+    new ApprovalService(actionService),
+  );
```
bferenc, line 35: Are these changes still needed after fixing the DI's annotations?
```suggestion
```

bferenc (changes requested):
Thanks, I had so much fun reviewing these changes! 😅 
Left a comment below, also it doesn't seem to reduce the build times by a lot (compared to other PR's CI build tasks) 😢

PR #137, opened 2025-02-07
feat: improved c-chain fees estimation

meeh0w:
_no ticket_

## Changes
* Updates EVM Module to take advantage of the new `eth_suggestPriceOptions` call on C-Chain
* Updates SDKs to use a more precise rounding technique for small amounts
* Updates naming of fee presets from `Normal` / `Fast` / `Instant` to `Slow` / `Normal` / `Fast`

## Testing
* Test fees for C-Chain and non-C-Chain networks
* Test both Fuji and Mainnet C-Chain (the first has the new call available, the 2nd does not) -- both should work (it should try to use `eth_suggestPriceOptions` and fall back to the regular fee fetching if it's not available)

## Checklist for the author


src/pages/SignTransaction/hooks/useGetTransaction.tsx
meeh0w, line 1: This was not being used.

bferenc (dismissed):
works as described 👍

bferenc (approved):

PR #142, opened 2025-02-10
feat: update vm-module versions to 1.4.2

csabbee:
## Description

This PR is to update the vm-module versions in order to utilize changes from **[here](https://github.com/ava-labs/vm-modules/pull/235)**

## Changes

* update module versions to `1.4.2`

## Screenshots:
Screenshot 2025-02-10 at 9 25 52

Screenshot 2025-02-10 at 9 25 59

## Checklist for the author


gergelylovas (approved):

meeh0w (commented):
I think the PR is missing the dependency update :) 

```json
"@avalabs/evm-module": "0.0.0-CP-9855-replace-glacier-with-d-20250213161239",
```

I think it would be beneficial if we look into ways to prevent malicious/suspicious tokens from leaking into the portfolio balance, before we switch away from Glacier.

I have a strong suspicion that `is_wallet` property on the result is what we're looking for, but I didn't find any docs to support it. All I know is that `ETHG` is the only token for me that has `is_wallet: false` and it's also the only one missing from [my DeBank portfolio view](https://debank.com/profile/0x05bcd07a055a950183d0b0cd88caa579fe58d2ea/?chain=eth&fromWalletApp=true) -- but they're also using a different endpoint (`/token/balance_list` -- which is not listed anywhere in their API docs).

See the screenshot: DOGE shows up on DeBank's portfolio view, even though it's `is_verified: false && is_wallet: true`. ETHG does not and it's `is_verified: false && is_wallet: false`.

image

image


Ooor, maybe it's `is_core`:

> `is_core:boolean` - Whether or not to show as a common token in the wallet.

csabbee:
Oh wow I actually managed to not commit the dependency change? O.o

meeh0w (approved):
Tested and works as expected (no Glacier calls for Ethereum, and also no scammy tokens from DeBank anymore).

src/background/services/seedless/SeedlessBtcSigner.ts
```diff
@@ -5,6 +5,7 @@ import {
 } from '@avalabs/core-wallets-sdk';
 import { SignerSession } from '@cubist-labs/cubesigner-sdk';
 import { Network, Psbt, SignerAsync, payments } from 'bitcoinjs-lib';
+import { TransactionInput, TransactionOutput } from 'bip174/src/lib/interfaces';
```
meeh0w, line 8: Hmm, `bip174` is not a direct dependency of the extension; it's only there because of other dependencies, so I'm not sure if we can rely on it to be there & `import` it like that.

Also, I'm pretty sure updating `@avalabs/avalanchejs` to the same version that's used by vm modules (`4.2.0-alpha.3`) will fix those typings issues. 🤔

meeh0w (approved):

PR #143, opened 2025-02-10
feat: cp-9885 autolock the extension after a given time

vvava:
## Description

[https://ava-labs.atlassian.net/browse/CP-9885](url)

## Changes

Removed keep-alive script and added alarms.

## Testing

Before that you should set the autolock time to a few seconds (e.g. 5/60 = 5 seconds) in the `src/background/services/lock/LockService.ts line 21`
Open the extension -> unlock it and close -> after 5 seconds the extension should be locked
It must not fire the alarm: 
 - if one or more extension window is open 
 - if the extension is locked by default

## Screenshots:

https://github.com/user-attachments/assets/07030a85-ae45-4797-89fa-02a0752605c8

## Checklist for the author


meeh0w (approved):
Nice! Clean! 👏 

Not for right now (maybe as a part of the redesign), but it gave me an idea that maybe we could add a 🔒 icon when extension is locked :)

Similar to 1Password:

image

vvava:
> Nice! Clean! 👏
> 
> Not for right now (maybe as a part of the redesign), but it gave me an idea that maybe we could add a 🔒 icon when extension is locked :)
> 
> Similar to 1Password:
> 
> image

I will collect and create some mvps for how we could improve a bit more the functionality of the extension and this was going to be a part of that, we are thinking about the same :)

bferenc (approved):

PR #129, opened 2025-01-30
chore: improve swap failure messaging & log them to Sentry

meeh0w:
## Changes
* Improves messaging of possible swap failures (avoid generic `Swap Failed` toast message)
* Logs failures to Sentry

## Testing
Make swap transactions fail. For example:
* block Paraswap in dev tools
* in two-step swaps, approve spend limit, revoke it and then approve the 2nd transaction
* modify code

## Screenshots:
image

## Checklist for the author
- [ ] I've covered new/modified business logic with Jest test cases.
- [x] I've tested the changes myself before sending it to code review and QA.

PR #146, opened 2025-02-11
feat: CP-9247 add toast to swap status

vvava:
## Description

[https://ava-labs.atlassian.net/browse/CP-9247](url)

## Changes

Move the toasters to the swap provider to handle the notifications.

## Testing

Go to swap -> approve a transaction and watch the toasters, it should be pending after the failed ir success message

## Screenshots:

https://github.com/user-attachments/assets/55d1866f-0c25-41ea-92f5-762be99cdd96

## Checklist for the author

### DIFF
--- a/src/contexts/SwapProvider/SwapProvider.test.tsx
+++ b/src/contexts/SwapProvider/SwapProvider.test.tsx
@@ -93,6 +93,8 @@ jest.mock('ethers');
 jest.mock('@avalabs/core-k2-components', () => ({
   toast: {
     success: jest.fn(),
+    loading: jest.fn(),
+    dismiss: jest.fn(),
   },
 }));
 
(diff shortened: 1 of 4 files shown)


meeh0w (dismissed):
👏

meeh0w (approved):

csabbee (commented):
This is just a question in general, is it deliberate that we are not displaying any toast message if for non native tokens the allowance check fails?
I see that we have test cases for such scenario(s), but the toast message is never triggered in those cases

csabbee:
> This is just a question in general, is it deliberate that we are not displaying any toast message if for non native tokens the allowance check fails? I see that we have test cases for such scenario(s), but the toast message is never triggered in those cases

Ah I see that it will be covered in this PR: https://github.com/ava-labs/core-extension/pull/129

csabbee (approved):

meeh0w (approved):

PR #151, opened 2025-02-19
fix: do not require unlocking for wallet selector

meeh0w:
## Description
* [CP-9949](https://ava-labs.atlassian.net/browse/CP-9949)

## Checklist for the author

### DIFF
--- a/src/contexts/WalletProvider.tsx
+++ b/src/contexts/WalletProvider.tsx
@@ -6,6 +6,7 @@ import {
   useState,
   useMemo,
 } from 'react';
+import { useRouteMatch } from 'react-router-dom';
 import { useConnectionContext } from './ConnectionProvider';
 import { filter, map } from 'rxjs';
 import { WalletDetails } from '@src/background/services/wallet/models';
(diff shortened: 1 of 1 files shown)


src/contexts/WalletProvider.tsx
```diff
@@ -175,7 +176,11 @@ export function WalletContextProvider({ children }: { children: any }) {
     [request],
   );
 
-  if (!isWalletLoading && isWalletLocked) {
+  // We do not require extension to be unlocked for wallet selection
+  const routeMatch = useRouteMatch('/approve/select-wallet');
```
meeh0w, line 180: Maybe it would be better to have a separate provider for locked/unlocked state (like a `LockProvider`), but I wasn't sure if it's worth it since it's just one route.

csabbee (approved):
🔥 🚒

src/contexts/WalletProvider.tsx
gergelylovas: Let's keep it like this for now. We can refactor when we have more

gergelylovas (approved):

PR #163, opened 2025-02-27
fix: resolving Bitcoin module for bitcoin_* requests

meeh0w:
* [CP-9900](https://ava-labs.atlassian.net/browse/CP-9900)

## Description
With the introduction of HyperVM support, we changed how `EVMProvider` passes the `scope` to the `ChainAgnosticProvider`. It used to simply pass in the `chainId`, but since that change, it started passing `eip155:<chain_id_hex>`.

This was still being resolved by `NetworkService.getNetwork(...)` (thanks to fake chain ids in non-EVM networks), but the request handler middleware was not looking at the resolved network, but instead at the `request.scope`, which none of the VM Modules recognized anymore.

## Changes
This change takes advantage of `ActiveNetworkMiddleware`, which can accurately resolve the network and pass it down to the `ModuleManger` instead of relying on `request.scope`.

## Testing
Perform a BTC -> BTC.b bridge from Core Web. One note: Core Web has a bug and does not recognize the local extension build ([PR that fixes it](https://github.com/ava-labs/core-web/pull/243/files)).

To test locally, change the `EVM_PROVIDER_NAME` from `"Core Dev"` to `"Core"` in `webpack.inpage.js`:
https://github.com/ava-labs/core-extension/blob/5c01c817af95455637ef46b64059e82a31eac0b9/webpack.inpage.js#L11

## Screenshots:

https://github.com/user-attachments/assets/02b2367e-3730-423e-8166-59eac17882a1

## Checklist for the author


src/utils/caipConversion.ts
```diff
@@ -12,14 +15,14 @@ export enum CaipNamespace {
 }
 
 export const BitcoinCaipId = {
-  [ChainId.BITCOIN]: `${CaipNamespace.BIP122}:000000000019d6689c085ae165831e93`,
-  [ChainId.BITCOIN_TESTNET]: `${CaipNamespace.BIP122}:000000000933ea01ad0ee984209779ba`,
+  [ChainId.BITCOIN]: BitcoinCaip2ChainId.MAINNET,
```
meeh0w, line 18: Switched to `@avalabs/core-chains-sdk` since that's much lighter than the Wallets SDK (and it would be included in the inpage script).

vvava (approved):

PR #164, opened 2025-02-27
feat: CP-9423 gasless screens

vvava:
## Description

Implement the gasless feature to the extension [https://ava-labs.atlassian.net/jira/software/c/projects/CP/boards/72?selectedIssue=CP-9423](url)

## Changes

- adding offscreen
- the GasStationService which handles the background work
- integrate the frontend part to the NetworkFeeProvider and to the corresponding pages

## Testing

Go to c-chain -> try any transaction -> you should see the Gasless Switch -> try with that and without that, check when the account has no native token to send a transaction etc.

## Screenshots:

**The Happy Path**

https://github.com/user-attachments/assets/9cc11e85-471b-442a-997d-fc2c906a9b90

**REtrying**

https://github.com/user-attachments/assets/47e63867-1a11-4830-b5eb-edb7e3abc111

**Error and turn gasless off**

https://github.com/user-attachments/assets/3787c8bf-ee8c-4f63-ae85-a0686358aac9

**When the address has no AVAX at all**

https://github.com/user-attachments/assets/871adf85-de14-4146-a67c-45094c712efc

**When only the gas fund is successful but the transaction is failing** (the transaction has been never called in this case because of the testing) -> you would never seen again the gasless switch until a successful transaction without gasless fund has been ended

https://github.com/user-attachments/assets/c536eae3-d572-4813-a4c7-ecd94f806d9e

**When the gasless solution is still calculating or something happened and we cannot get the soltuion hex back, we notify the user**

https://github.com/user-attachments/assets/9a20fc71-d009-46cf-8bb4-7b6c67dc2dc3

## Checklist for the author


src/pages/ApproveAction/GenericApprovalScreen.tsx
```diff
@@ -69,34 +81,179 @@ export function GenericApprovalScreen() {
...
+    getGaslessEligibility,
+    getNetwork,
+    isGaslessEligible,
+    setIsGaslessEligible,
+    setIsGaslessOn,
+    signingData,
+  ]);
+
+  useEffect(() => {
+    createGaslessOffscreen();
```
gergelylovas, line 152: Why does the frontend need to control this? Shouldn't this be fully automatic from the backend? What happens if one approval screen opens an offscreen and the other approval window closes it immediately?

src/pages/ApproveAction/GenericApprovalScreen.tsx
```diff
@@ -69,34 +81,179 @@ export function GenericApprovalScreen() {
...
+    setIsGaslessOn,
+    isFundProcessReady,
+    fundTxHex,
+    fundTxDoNotRertyError,
+    setGaslessDefaultValues,
+    createGaslessOffscreen,
+    closeGaslessOffscreen,
+    solutionHex,
+    isGaslessFundStarted,
+    setIsGaslessFundStarted,
```
gergelylovas, line 99: Seems like there are multiple `is**` state variables which intend to store the current state of the funding process. 
However many of these cannot be true at the same time or being set to true implies that the other values are already true. 
Suggestion: use an enum instead to store the current status of the gasless flow. E.g:
Gasless: READY, ELIGIBLE, FUNDING_IN_PROGRESS, FUNDED, ERROR

src/contexts/utils/getCurrencyFormatter.ts
```diff
@@ -44,3 +44,13 @@ const modifyFractionNumber = (amount: number) => {
...
+
+const parseScientificNotation = (amount: string) => {
+  if (amount.toString().includes('e') && amount.toString().includes('-')) {
+    return ['0', '00'];
+  }
+  if (amount.toString().includes('e') && amount.toString().includes('+')) {
+    return ['Infinity', 'Infinity'];
+  }
+  return amount.toString().split('.');
+};
```
gergelylovas, line 56: Can you please add some comments and tests to this to explain what the goal is here and why is it okay to round to 0.00 and Inifinity.Infinity?

src/background/connections/offscreenConnection/OffscreenConnectionController.ts
```diff
@@ -0,0 +1,203 @@
...
+      );
+    }
+    this.connection?.postMessage(message);
+  }
+
+  private async onMessage(requestJSON: string) {
+    if (!this.pipeline || !this.connection) {
+      throw Error('OffscreenConnectionController is not connected to a port');
+    }
+    await this.hasOffscreenDocument('offscreen.html');
```
gergelylovas, line 104: Is this needed here?

src/background/services/gasless/GasStationService.ts
```diff
@@ -0,0 +1,195 @@
...
+      }
+    }
+  }
+
+  setDefaultStateValues() {
+    this.isFundProcessReady.dispatch(false);
+    this.fundTxHex.dispatch('');
+    this.fundTxDoNotRertyError.dispatch(false);
+    this.solutionHex.dispatch('');
+    this.challengeHex.dispatch('');
```
gergelylovas, line 117: Whats the reason for having a separate event for each property of the state?

src/contexts/NetworkFeeProvider.tsx
```diff
@@ -13,22 +15,87 @@ import { useConnectionContext } from '@src/contexts/ConnectionProvider';
...
   const [iteration, setIteration] = useState(0);
+  const [challengeHex, setChallengeHex] = useState('');
+  const [solutionHex, setSolutionHex] = useState('');
+  const [isGaslessOn, setIsGaslessOn] = useState(false);
+  const [isGaslessEligible, setIsGaslessEligible] = useState<boolean | null>(
+    null,
+  );
+  const [fundTxHex, setFundTxHex] = useState('');
+  const { featureFlags } = useFeatureFlagContext();
+  const [gaslessPhase, setGaslessPhase] = useState<GaslessPhase | null>(null);
```
gergelylovas, line 98: Instead of allowing null, it would be nicer to have a `GaslessPhase.READY` value. This would mean no null checks and the state is more explicit.

src/contexts/NetworkFeeProvider.tsx
```diff
@@ -13,22 +15,87 @@ import { useConnectionContext } from '@src/contexts/ConnectionProvider';
...
+  gaslessFundTx: ({
+    data,
+    fromAddress,
+  }: {
+    data: TransactionRequest;
+    fromAddress: string;
+  }) => Promise<string | undefined>;
+  isGaslessOn: boolean;
+  setIsGaslessOn: Dispatch<SetStateAction<boolean>>;
+  isGaslessEligible: boolean | null;
```
gergelylovas, line 42: The  `isGaslessEligible` and `isGaslessOn` flags can also be folded into `gaslessPhase`. Since the gasless cannot be FUNDED if its not ON and it cannot be ON if it's not ELIGIBLE.

gergelylovas (changes requested):
Nice work, getting there. Some more small tweaks

src/background/services/gasless/handlers/fundTx.ts
```diff
@@ -0,0 +1,45 @@
+import { ExtensionRequest } from '@src/background/connections/extensionConnection/models';
+import { ExtensionRequestHandler } from '@src/background/connections/models';
+import { injectable } from 'tsyringe';
+import { GasStationService } from '../GasStationService';
+import { TransactionRequest } from 'ethers';
+// import { DEFERRED_RESPONSE } from '@src/background/connections/middlewares/models';
```
gergelylovas, line 6: Nit: please remove

gergelylovas (approved):
Nice one! Much cleaner. 
Let's do the tests as part of the follow up ticket.

csabbee (approved):

bferenc (approved):
Works great. Tried sending AVAX / ERC20, swap with both cases even on accounts with 0 AVAX 👍

PR #168, opened 2025-03-12
refactor: how we store permissions

meeh0w:
⚠️ Based on #133 

## Description
* [CP-10046](https://ava-labs.atlassian.net/browse/CP-10046)

## Changes
* Parts of this PR exist solely because of SDK updates and the typings I had to fix as a result.
* It updates the `permissions` storage schema so that it accepts non-EVM addresses too.
* It introduces `accountsChangedCA` event that emits not only the address, but also the VM it is derived for, so that injected providers can choose which addresses to pass to the dApps.
    * Ideally, it would replace `accountsChanged` completely and update the `EVMProvider` from `@avalabs/evm-module`, to listen to this even too, but I didn't want to do this within this PR.

## Testing
You gotta test the migration and connecting to dApps.

* Onboard on `main` (or any branch without these changes here)
* Switch to this branch and everything should still work as usual. This lays foundation for https://github.com/ava-labs/core-extension/pull/170
* Connect to some dApps, it should work as usual (try both when extension is locked and unlocked)
* You should be able to revoke permissions as usual (via Settings -> Security and Privacy -> Connected Sites or via the extension header)

## Checklist for the author
- [x] I've covered new/modified business logic with Jest test cases.
- [x] I've tested the changes myself before sending it to code review and QA.

PR #169, opened 2025-03-12
feat: Solana balances, signing, sending and activity history

meeh0w:
* Based on #168 

## Description
* [CP-9731](https://ava-labs.atlassian.net/browse/CP-9731)

## TODO
- [x] Derive Solana address for users who switch to the Solana network

## Changes
* Starts fetching Solana networks from Contentful (fetching networks via `/tokenlist?includeSolana` instead of just `/tokenlist`)
* Imports the `SvmModule` into Extension
* Implements send flow for Solana

## Testing
* Do a clean onboarding (to derive the Solana address -- I'll make it automatic when switching to Solana in a different task)
* Switch to Solana network
* Verify you see the correct address (i.e. compare with Phantom or other Solana-capable wallet)
* Send some funds to your other account. Test 2 flows:
    1. where target account DOES NOT have any funds yet (for this case needs, there is a minimum amount that can be sent which Extension should validate and show you if you try to send too little)
    2. where target account DOES HAVE some funds already
* Verify your transactions show in the activity tab

## Screenshots:
⚠️ The wrong transaction date is getting fixed in a VM Modules PR.

https://github.com/user-attachments/assets/6376d52d-377f-4f8c-8810-d16dbc1d009c

## Checklist for the author


src/utils/caipConversion.ts
```diff
@@ -25,6 +33,8 @@ export const AvaxCaipId = {
 export const getNetworkCaipId = (network: PartialBy<Network, 'caipId'>) => {
   if (network.caipId) {
     return network.caipId;
+  } else if (network.caip2Id) {
+    return network.caip2Id;
   }
```
gergelylovas, line 38: Why do we have both `caipId` and `caip2Id`? Aren't the two the same?

src/background/services/wallet/WalletService.ts
```diff
@@ -530,12 +550,30 @@ export class WalletService implements OnUnlock {
...
+          tx.data,
+          (await getProviderForNetwork(network)) as SolanaProvider,
+        );
+
+        return {
+          signedTx,
+        };
+      }
+
+      throw new Error('Unable to find a proper signer');
```
gergelylovas, line 574: Nit: We should throw early
```suggestion
      if (!(wallet instanceof SolanaSigner)) {
        throw new Error('Unable to find a proper signer');
      }
      const signedTx = await wallet.signTx(
        tx.data,
        (await getProviderForNetwork(network)) as SolanaProvider,
      );

      return {
        signedTx,
      };
```

src/pages/Send/Send.tsx
```diff
@@ -232,6 +226,22 @@ export function SendPage() {
             onApproved={onApproved}
           />
         )}
+      {!isLoading &&
+        network.vmName === NetworkVMType.SVM &&
+        isSvmCapableAccount(active) && (
+          <SendSVM
+            network={network}
+            fromAddress={fromAddress}
+            maxFee={0n} // TODO: fix
```
gergelylovas, line 235: Do we need to fix this?

gergelylovas (changes requested):

src/utils/caipConversion.ts
meeh0w: They are...when I was naming the field in Contentful, I made the mistake of naming it `caip2Id` and it propagated through the Proxy API and `@avalabs/core-chains-sdk`.

So far the `caipId` only lived in the extension and was populated programatically by the `decorateWithCaipId(...)` util function:

https://github.com/ava-labs/core-extension/blob/5209e15ce2860ddcf2470617db84ea2032ebf019/src/background/services/network/models.ts#L37

https://github.com/ava-labs/core-extension/blob/876fd7dc0974f6fb16719844ff38659f862afc9a/src/utils/caipConversion.ts#L121-L126

I didn't want to blow these PRs with unnecessary refactorings, so my plan was to just go with it in until the rollout and later get rid of the old field (rename it to `caip2Id` everywhere, since it's also a more fitting name).

src/pages/Send/Send.tsx
meeh0w: Not really, since we're not showing the fee widget for Solana at the moment.

But I do need to take the fixed 5000 Lamports fee into consideration, so thanks for catching this and reminding me of that 💡

src/pages/Send/components/AddressDropdownListMyAccounts.tsx
```diff
@@ -49,17 +50,20 @@ export const AddressDropdownListMyAccounts = ({
   const { t } = useTranslation();
   const { network } = useNetworkContext();
   const useBtcAddress = isBitcoin(network);
+  const useSvmAddress = network && isSolanaNetwork(network);
```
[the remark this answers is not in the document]
meeh0w: Definitely. Overall, I will want to refactor how we handle `Contact`s and own addresses in a separate ticket, since the whole thing with `addressC`, `addressBTC`, `addressHVM`, etc. is just horrible imho.

I think having a dictionary field like `addresses: Record<NetworkVMType, string>` field would be much much better. This also goes for the `Account` model.

src/background/services/wallet/WalletService.ts
meeh0w: Fixed here: 8bb9887c0

csabbee (dismissed):
I didn't have enough balance to send SOL, but used a faucet to receive some, it showed up on the activity tab.
Nice job! 🔥

src/background/services/wallet/WalletService.ts
```diff
@@ -710,6 +748,7 @@ export class WalletService implements OnUnlock {
       evm: evmPub?.key,
       xp: avmPub?.key,
       ed25519: hvmPub?.key,
+      svm: secrets.account.addressSVM,
```
bferenc, line 751: Is it true? Don't we have to decode the address?

src/background/vmModules/ApprovalController.ts
```diff
@@ -344,6 +344,14 @@ export class ApprovalController implements BatchApprovalController {
         return await this.#walletService.sign(
           signingData.data,
           network,
+
```
bferenc, line 347: 🔥 
```suggestion
```

src/background/services/wallet/WalletService.ts
meeh0w: It's technically the same value, but maybe we should indeed hex-encode it first. I'm unsure, because no one is using this value at the moment (Solana dApps fetch the public key through a different API).

bferenc: Hmm my thinking was if a client uses the `avalanche_getAccountPubKey` API, they would expect to get back actual public keys and not their encoded values 🤔 
Anyways, I don't mind leaving it as is now. 👍

bferenc (commented):
Balance and activity list shows up correctly, but as I go to the `Send` page it gets stuck in a loading state (logs are empty)
image
image

Is it just me?

bferenc (approved):
thansk for the quick fix! sending, signing also works ✅

bferenc (dismissed):

bferenc (dismissed):

csabbee (approved):

csabbee (approved):
🔥

PR #199, opened 2025-03-28
feat: CP-10056 allow chainId

vvava:
## Description

[CP-10056](https://ava-labs.atlassian.net/browse/CP-10056)

## Changes

Check the chainId besides the caipId.

## Testing

Go to a custom network and try it out -> you can modify the condition around here: https://vscode.dev/github/ava-labs/core-extension/blob/feat/CP-10056_custom-netowrk-chainid-support/src/background/services/wallet/handlers/eth_sendTransaction/utils/getTargetNetworkForTx.ts#L48

## Checklist for the author

### DIFF
--- a/src/background/services/wallet/handlers/eth_sendTransaction/utils/getTargetNetworkForTx.ts
+++ b/src/background/services/wallet/handlers/eth_sendTransaction/utils/getTargetNetworkForTx.ts
@@ -42,19 +42,6 @@ const getTargetNetworkForTx = async (
     });
   }
   assertSameEnvironment(network, uiActiveNetwork);
-  const isCustomNetwork = network.chainId in networkService.customNetworks;
-
-  // Only allow custom networks if they are also the active network for the dApp
-  if (
-    isCustomNetwork &&
-    network.caipId !== activeScope &&
-    network.chainId !== uiActiveNetwork?.chainId
-  ) {
-    throw ethErrors.rpc.invalidParams({
-      message: 'Provided custom network is not supported',
-      data: { chainId },
-    });
-  }
 
   return network;
 };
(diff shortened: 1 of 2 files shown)


src/background/services/wallet/handlers/eth_sendTransaction/utils/getTargetNetworkForTx.ts
```diff
@@ -42,13 +42,16 @@ const getTargetNetworkForTx = async (
...
   assertSameEnvironment(network, uiActiveNetwork);
-
   const isCustomNetwork = network.chainId in networkService.customNetworks;
 
   // Only allow custom networks if they are also the active network for the dApp
-  if (isCustomNetwork && network.caipId !== activeScope) {
+  if (
+    isCustomNetwork &&
+    network.caipId !== activeScope &&
+    network.chainId !== uiActiveNetwork?.chainId
```
meeh0w, line 51: Something feels wrong here, probably there's a misunderstanding 🤔 

We were supposed to _loosen_ the restriction, meanwhile you're _adding another check_ and not removing any.

The ticket says:
>  I would like to work on my new L1 with as little friction as possible. Currently constant chain switching is required between the primary network and the new L1.

This change here does not change it, chain switching will still be required.

Reading the acceptance criteria, from now on we should only require that the custom network is in the same environment (testnet / mainnet) as the extension's UI.

So I'm pretty sure we should remove this whole `if` instead, as the environment being the same is asserted in line `44`.

meeh0w (changes requested):

meeh0w (approved):

csabbee (approved):

PR #205, opened 2025-03-31
fix: xp-chains regressions

meeh0w:
## Changes
* Fixes missing `CoreEth` address (by updating the VM Modules), which in turn fixes Core Web connection & X/P-chain send flows
* Updates assertion for extended public keys being present in the `WalletService` (it was asserted too early).

## Checklist for the author


src/background/services/wallet/WalletService.ts
```diff
@@ -915,34 +915,32 @@ export class WalletService implements OnUnlock {
       return [];
     }
 
-    assertPresent(indices[0], SecretsError.NoAccountIndex);
```
meeh0w, line 918: This assertion was too early, this sometimes gets called this without any indices passed

vvava (approved):

bferenc (approved):

gergelylovas (approved):

PR #206, opened 2025-03-31
fix: set minimum slippage

meeh0w:
## Changes
Sets minimum slippage for Swaps to 0.1%

## Checklist for the author

### DIFF
--- a/src/pages/Swap/components/TransactionDetails.tsx
+++ b/src/pages/Swap/components/TransactionDetails.tsx
@@ -13,7 +13,11 @@ import {
   InfoCircleIcon,
 } from '@avalabs/core-k2-components';
 import { PARASWAP_PARTNER_FEE_BPS } from '@src/contexts/SwapProvider/constants';
-import { formatBasisPointsToPercentage } from '../utils';
+import {
+  formatBasisPointsToPercentage,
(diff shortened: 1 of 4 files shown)


bferenc (dismissed):

src/pages/Swap/components/TransactionDetails.tsx
```diff
@@ -142,19 +146,26 @@ export function TransactionDetails({
...
                     max: 100,
+                    step: 0.1,
                   },
                 }}
                 onChange={(e) => {
                   const value = e.target.value;
-                  const isValid = isSlippageValid(value);
-                  if (!isValid) {
-                    return;
+                  if (isSlippageValid(value)) {
```
gergelylovas, line 156: I think this change handler makes the UX a bit jarring. When 0.1 is reached, there is no way to delete and enter a number. Also when I enter a `0.` to enter let's say `0.5` it jumps to `0.1` and it's not very user friendly to get it to the desired number.

https://github.com/user-attachments/assets/93a743fa-54fe-4195-a161-671b895daf0f

gergelylovas (changes requested):

gergelylovas (approved):

PR #208, opened 2025-03-31
fix: explorer URLs

meeh0w:
## Changes
* Respect the `pathname` coming from `network.explorerUrl`

## Checklist for the author

### DIFF
--- a/src/contexts/SwapProvider/SwapProvider.test.tsx
+++ b/src/contexts/SwapProvider/SwapProvider.test.tsx
@@ -553,7 +553,7 @@ describe('contexts/SwapProvider', () => {
       });
     });
 
-    fdescribe('when swapping non-native token', () => {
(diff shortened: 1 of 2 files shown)


src/contexts/SwapProvider/SwapProvider.test.tsx
```diff
@@ -553,7 +553,7 @@ describe('contexts/SwapProvider', () => {
       });
     });
 
-    fdescribe('when swapping non-native token', () => {
+    describe('when swapping non-native token', () => {
```
meeh0w, line 556: Also found a bunch of skipped tests, fixing that.

gergelylovas (approved):
tested on swap. Works. 
on Ethereum we have a `//` in the url but seems to be working alright

vvava (approved):

PR #213, opened 2025-04-03
feat: CP-10233 filter toToken list

vvava:
## Description

[CP-10233](https://ava-labs.atlassian.net/browse/CP-10233)

## Changes

Filtering the To token in Swap.

## Testing

Go to swap -> check the From selected token should not appear in the To tokens list.

## Screenshots:

https://github.com/user-attachments/assets/c9a44202-9d73-473c-b21e-17dcd7a2c1e2

## Checklist for the author

### DIFF
--- a/src/pages/Swap/Swap.tsx
+++ b/src/pages/Swap/Swap.tsx
@@ -79,14 +79,6 @@ export function Swap() {
     forceShowTokensWithoutBalances: true,
     disallowedAssets: DISALLOWED_SWAP_ASSETS,
   });
-  const allSwappableTokens = useMemo(
-    () =>
-      allTokensOnNetwork.filter(
-        (token) =>
-          token.type === TokenType.ERC20 || token.type === TokenType.NATIVE,
-      ),
-    [allTokensOnNetwork],
-  );
   const {
     accounts: { active: activeAccount },
   } = useAccountsContext();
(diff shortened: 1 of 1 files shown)


bferenc (commented):
This does the trick, but I'm not sure if this is what we want 🤷 

This way I can still select the same token for both "from" and "to" but they then change places and "From" gets replaced by a random token from top of the list.

Question: shouldn't we do the same what mobile does?

src/pages/Swap/Swap.tsx
```diff
@@ -314,7 +314,11 @@ export function Swap() {
               setIsFromTokenSelectOpen(!isFromTokenSelectOpen);
               setIsToTokenSelectOpen(false);
             }}
-            tokensList={tokensWBalances}
+            tokensList={tokensWBalances.filter(
+              (token) =>
+                token.name !== selectedToToken?.name ||
+                token.symbol !== selectedToToken?.symbol,
+            )}
```
meeh0w, line 321: What Feri said + I'm pretty sure calling `.filter()` here (in the DOM tree) will make it such that `<TokenSelect>`s will always be re-rendered, even if `tokensWBalances` and `allSwappableTokens` don't change.

vvava: I'm not sure how the mobile wants to do with this but you won't able to pick the selected token from the other list and that would be the point of the ticket (resetting and / or setting default values - as we can see in the linked video - makes no sense here since we have a pair of tokens by default).

meeh0w (approved):

bferenc (approved):

PR #215, opened 2025-04-07
[CP-10261] check active account and network only when needed

bferenc:
## Description
When we check for feature availability we should only care about correctly set `activeAccount` and `activeNetwork` when it's actually necessary for the feature.

## Changes
`isFunctionAvailable` no longer defaults to `false` until `activeAccount` and `activeNetwork` are loaded. If any of them is required for the feature availability check `checkIsFunctionAvailable` will validate them.

## Testing
- initiate any X/P transaction from Core web (e.g.: cross-chain transfer)
- verify the "This is currently unavailable" warning does not appear

## Checklist for the author

### DIFF
--- a/src/hooks/useIsFunctionAvailable.ts
+++ b/src/hooks/useIsFunctionAvailable.ts
@@ -195,6 +195,7 @@ export const useIsFunctionAvailable = (
   const {
     accounts: { active },
   } = useAccountsContext();
+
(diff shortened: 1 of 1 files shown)


meeh0w (approved):
I hereby confirm I don't see the `function unavailable` message.


<small><sub>*but I also didn't see it before :)*</sub></small>

PR #219, opened 2025-04-10
feat: CP-10098 store results, handle multiple call

vvava:
## Description

We had false result for eligibility checks.

## Changes

Store the result when the call only for the chainId (in this case we won't call the sdk for checking the eligibility all the time with only the chainId).
Store the data of the calls so we can return in the right order each of them.

## Testing

Test the eligibility is working on different chains.

## Screenshots:

https://github.com/user-attachments/assets/c7110dbb-8aa3-4aa3-b7a5-a29cb1f5ed02

## Checklist for the author


vvava:
something is still wrong, I will reopen

PR #231, opened 2025-04-17
fix: missing encryption key error

meeh0w:
## Description
We were attempting to load the wallet details as soon as `<WalletProvider>` was rendered, which also happens on the lock screen.

## Changes
* Only attempts to load the wallet details after extension gets unlocked.

## Testing
* Open dev tools
* Lock the extension
* You should not see the `Encryption key missing` error message anymore

## Checklist for the author


csabbee (approved):
🔥 🚒

vvava (approved):

PR #237, opened 2025-04-22
[CP-10023] push notifs

bferenc:
## Description
Enables push notifications for the extension for the following categories:
- balance changes
- news (price alerts, product announcements, offers and promotions, market news)

## Changes
- Adds a `Notifications` page under `Advanced` settings where users can subscribe to and unsubscribe from the aforementioned categories
- Adds device registration. The device's ID is persisted in the unencrypted storage and the corresponding endpoint is always updated whenever the FCM token changes (e.g.: on version update)
- Adds subscription management logic

## Testing
- On the first device registration, you should get a device ARN
- On subsequent startups the `push/register` requests should contain your device ARN
- Unlock the extension, navigate to Advanced -> Notifications
- Subscribe / unsubscribe as you wish
- Confirm that you receive a balance change notification whenever you transact with one of your accounts on C-chain (should work on both mainnet and fuji))
- Confirm that you receive news notifications by creating `Push Notification` entries on contentful (Core space, staging env)
- Confirm that you do not receive notifications when you are not subscribed to a given category

## Checklist for the author


src/contexts/NotificationsProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+        params: notificationType,
+      }).finally(() => syncSubscriptions());
+    },
+    [request, syncSubscriptions],
+  );
+
+  useEffect(() => {
+    syncSubscriptions().catch(
+      // noop
+      () => {},
```
meeh0w, line 84: I think instead of swallowing every error here, we should check if the extension is unlocked first (either here, or in the `GetNotificationSubscriptions` handler).

Then instead of a `noop`, we could actually log the error here so we know if something ever goes wrong.

src/components/settings/pages/Notifications.tsx
```diff
@@ -0,0 +1,138 @@
...
+      setIsLoading(true);
+
+      handler()
+        .catch(() => toast.error(t('Failed to update notification settings')))
+        .finally(() => setIsLoading(false));
+
+      capture('NotificationSettingsUpdated', {
+        type: notificationType,
+        enabled: isChecked,
+      });
```
meeh0w, line 76: Should this happen in `.then` or `.catch` above?

meeh0w (approved):
Looks great! I only left some minor comments 👏

src/components/settings/pages/Notifications.tsx
bferenc: thanks! 
[ec97a26](https://github.com/ava-labs/core-extension/pull/237/commits/ec97a2634aa4087d1ab8039087a4085aa09fc46a)

src/contexts/NotificationsProvider.tsx
bferenc: here: [58de4c5](https://github.com/ava-labs/core-extension/pull/237/commits/58de4c5aa8be047710311283f376ed0f04e1ee8d)

csabbee (approved):
tested, working:)

PR #241, opened 2025-04-24
fix: disable solana dapp connection

meeh0w:
Since Solana support will not be publicly released until it's also supported by Core Mobile, we need to disable the inpage provider (feature flags are not available in the inpage script).

When mobile will be near the release, we can add a feature flag check on the connection screen, but for now let's just not display `Core` as an option on the dapps at all to avoid confusion.


csabbee (approved):
🔥

gergelylovas (approved):

PR #243, opened 2025-04-30
feat:  CP-10438 set default gas fee modifier to fast

vvava:
## Changes

Set the gas fee modifier to fast on c-chain and Fuji.

## Testing

e.g. Go to send and check the default modifier value -> try it on other networks

## Screenshots:

updated: 

https://github.com/user-attachments/assets/ba711c61-75be-4b17-b175-0e4371d64d2c

## Checklist for the author

### DIFF
--- a/src/pages/ApproveAction/hooks/useFeeCustomizer.tsx
+++ b/src/pages/ApproveAction/hooks/useFeeCustomizer.tsx
@@ -28,7 +28,7 @@ import {
   MultiTxAction,
   isBatchApprovalAction,
 } from '@src/background/services/actions/models';
-import { ChainId } from '@avalabs/core-chains-sdk';
+import { isAvalancheNetwork } from '@src/background/services/network/utils/isAvalancheNetwork';
 
 const getInitialFeeRate = (
   data?: SigningData | MultiTxFeeData,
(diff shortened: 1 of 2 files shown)


src/pages/ApproveAction/hooks/useFeeCustomizer.tsx
```diff
@@ -85,6 +86,16 @@ export function useFeeCustomizer({
   const [gasFeeModifier, setGasFeeModifier] = useState<GasFeeModifier>(
     GasFeeModifier.SLOW,
   );
+
+  useEffect(() => {
+    if (
+      network?.chainId === ChainId.AVALANCHE_MAINNET_ID ||
+      network?.chainId === ChainId.AVALANCHE_DEVNET_ID
+    ) {
```
meeh0w, line 94: I believe it should be `AVALANCHE_TESTNET_ID` instead. Or better use the helper:

```suggestion
    if (network && isAvalancheNetwork(network)) {
```

meeh0w (changes requested):
I'm not really sure, but should we also change this?

https://github.com/ava-labs/core-extension/blob/84bf24015e7729d771b28ce982277024c23cf0f5/src/components/common/CustomFees.tsx#L270-L280

to something like:

```ts
else if (isAvalancheNetwork(network)) {
    updateGasFee(GasFeeModifier.FAST);
}
```

It may be that `useFeeCustomizer` takes care of that already, in which case just ignore :)

meeh0w (approved):

PR #248, opened 2025-05-07
fix: cp-10546 swap tokens list

vvava:
## Description

We want to let the user swap to a token which is not currently owned by the user.

Merge [https://github.com/ava-labs/core-extension/pull/247](url) first.

## Changes

Add the list to the AI model.

Side note: It will let the user spend more token than it is in the wallet. 

## Testing

Try to swap to a new token.

## Screenshots:

https://github.com/user-attachments/assets/be7a6611-5ff7-4b93-bffd-912cc01cebb6

## Checklist for the author

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ typings/
 .idea
 .idea/*
 
+__dist
 dist
 builds
 .DS_Store
--- a/package.json
+++ b/package.json
@@ -5,11 +5,18 @@
   "private": true,
   "scripts": {
     "start": "yarn dev",
+    "rsbuild:build:inpage": "rsbuild build --config rsbuild.inpage.ts",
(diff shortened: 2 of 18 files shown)


meeh0w (approved):

csabbee (commented):
Tested, managed to work, but initially it couldn't recognize if I was trying to swap more tokens than I had, than it started working 🤷 
The other thing is that the recipient is spelled wrong in the `model.ts` and in `Prompt.tsx`: `recepient`
I don't think this should cause much issue, but who knows

meeh0w (approved):

csabbee (approved):
Tested, managed to swap back and forth 🔥

meeh0w (approved):

PR #252, opened 2025-05-08
fix: handle the error ai does not respond with text but status ok

vvava:
## Description

When the AI does not parse the function call correctly (or whatever the reason is) it responds with a status 200 but the text is empty.

## Changes

Let the user know something happened about the request.

## Testing

Go to AI -> the way I could see this error when I tried to get my balance with ```balance, token balance etc``` commands, try these

## Screenshots:

image

## Checklist for the author


bferenc (approved):

csabbee (approved):
nice, managed to trigger the error, working like a charm!

PR #318, opened 2025-06-30
feat(next-gen): Account Management

bartosz-the-coder-at-avalabs:
## Description

https://ava-labs.atlassian.net/browse/CP-10789
https://www.figma.com/design/kksuPP4PnRIFDx9Sm3544F/Core-Extension-Redesign-2025?node-id=1061-31256&m=dev
https://www.figma.com/design/kksuPP4PnRIFDx9Sm3544F/Core-Extension-Redesign-2025?node-id=913-52285&m=dev

## Changes

- Adds account rename feature
- Adds account remove feature
- Improves keyboard navigation
- Optimizes useKeyboardShortcuts hook

## Screenshots:

https://github.com/user-attachments/assets/2e05cb09-47eb-48bf-b25e-eda3eb405a71

## Checklist for the author


apps/next/src/pages/AccountManagement/components/AccountDetails/components/RemoveAccount.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <Typography variant="body1">
+          {t('Deleting this account is permanent and cannot be undone')}
+        </Typography>
+      </Stack>
+      <ActionButtons
+        top={{
+          label: t('Yes, delete'),
+          onClick: onDelete,
+          color: 'secondary',
+          panic: true,
```
csabbee, line 39: Panic 😄

csabbee (dismissed):
Nicely done! Tested, working as described

apps/next/src/pages/AccountManagement/components/NavigateBackContext.tsx
```diff
@@ -0,0 +1,49 @@
+import { createContext, FC, ReactNode, use, useMemo } from 'react';
+import { useHistory } from 'react-router-dom';
+
+type UnregisterFn = () => void;
+type NavigateBackContext = {
+  register: (callback: VoidFunction) => UnregisterFn;
+  goBack: VoidFunction;
```
meeh0w, line 7: TIL about `VoidFunction` 💡

apps/next/src/pages/AccountManagement/components/NavigateBackContext.tsx
meeh0w, line 1: This looks useful, should we move it higher up the directory tree so it can be used by other pages too?

apps/next/src/pages/AccountManagement/components/NavigateBackContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const context = useMemo<NavigateBackContext>(() => {
+    let onGoBack: VoidFunction | undefined;
+    return {
+      register: (goBackFn: VoidFunction) => {
+        onGoBack = goBackFn;
+        return () => {
+          onGoBack = undefined;
+        };
+      },
+      goBack: () => (onGoBack ?? goBack)(),
```
meeh0w, line 30: 🤔 Does this not work?

```suggestion
      goBack: onGoBack ?? goBack,
```

meeh0w (approved):
I have a few comments, but they can be follow-ups since the account creation & bulk removal is missing from this one :)

1. (not part of this PR, but still) The scrollbar looks a bit yucky when in the floating window:
image

UX note (something we should bring up with the UX team) - the scrollable area does not make sense to me, right now we're displaying the `Your account's private key is a...` note when the `Show private key` button is not even visible.

2. The slide-up animation for removal/renaming is awesome, but a scrollbar appears which makes it look bad:

https://github.com/user-attachments/assets/78fe55fb-551b-4e32-9009-dc5517c6a0be

3. Last note, removing the last account should not be possible (disable the button):

https://github.com/user-attachments/assets/4cdc0c6d-9e07-486d-bf89-03947d4ab9fb

apps/next/src/pages/AccountManagement/components/NavigateBackContext.tsx
bartosz-the-coder-at-avalabs: Nope. 
The point of this implementation is to lazy evaluate the actual `goBack` function
Since `register` function doesn't trigger re-render, the component which uses `context.goBack` will always use `history.goBack` function, because at the point of rendering there's a big change that `onGoBack` will be undefined

bartosz-the-coder-at-avalabs:
@meeh0w Points 2 & 3 have been resolved. 
Point 1 (scrollbars) will be address in as a separate task.

csabbee (approved):

PR #325, opened 2025-07-03
[Next-Gen] Show Private Key

bartosz-the-coder-at-avalabs:
## Description
Introduces Show Private Key screen

https://www.figma.com/design/kksuPP4PnRIFDx9Sm3544F/Core-Extension-Redesign-2025?node-id=1061-31607&m=dev
https://ava-labs.atlassian.net/browse/CP-10896

## Screenshots:

image
Screenshot 2025-07-03 at 17 42 24

## Checklist for the author


apps/next/src/pages/AccountManagement/components/ShowPrivateKey/components/EnterPassword.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Button
+        ref={revealButtonRef}
+        variant="contained"
+        color="primary"
+        size="small"
+        fullWidth
+        loading={isLoading}
+        disabled={!password || isLoading}
+        onClick={() =>
+          revealKey(password, chain, type, index, id).then(onAuthenticated)
```
csabbee, line 83: I know here it's pretty safe, because almost all the parameters are of different type, but how about configuring eslint that over 3 or 4 params, you need to pass in an object (and can use destructuring on the other end).
I run into issues in the pass when I was mixing up the order of params and was trying to hunt down bugs caused by that 😅 
\#foodForThought

csabbee (approved):
tested, working as described. Left a comment which is just a minor thing and more of a discussion starter for later

apps/next/src/pages/AccountManagement/components/ShowPrivateKey/components/EnterPassword.tsx
bartosz-the-coder-at-avalabs: I was thinking about this as well.
Let's chat about this on a team meeting and decide which way is most preferable for all of us 😄

meeh0w (approved):

PR #329, opened 2025-07-07
fix: introduce read-write lock for AccountsService

meeh0w:
* [CP-11370](https://ava-labs.atlassian.net/browse/CP-11370)

## Description
When switching from testnet to mainnet (or vice-versa), `AccountsService` recalculates all the account addresses asynchronously.
If a request for accounts (like `avalanche_getAccounts`) comes in while the recalculation is still in progress, `AccountsService` may serve stale addresses (i.e. `fuji` addresses when we're already on Mainnet, or `avax...` addresses when we're already on Testnet).

## Changes
* This PR introduces a Read-Write Lock for `AccountsService`. Direct access to `.accounts` or `.activeAccount` properties is removed and instead `.getAccounts()` and `.getActiveAccount()` methods should be used.
    * The concurrency classes are not written by me. I was looking for some stable and still maintained NPM libs, but didn't find anything promising (I found some that are no longer maintained, with latest contributions dating 7 years...). Instead, I used the source code provided in [this article](https://medium.com/@artemkhrenov/advanced-concurrency-patterns-in-javascript-semaphore-mutex-read-write-lock-deadlock-prevention-79e8bffb5b81), which I liked very much for its simplicity. The author mentions "full implementation with test cases" at the very bottom, but there is no link to the repo/package 😞 

* The `AccountsService` now:
    * may have many concurrent readers, as long as no write operation is in progress.
    * may only have one active writer at any given time
    * if a write operation is in progress, readers are queued up

## Testing
1. Connect to the new Core Web:
    * https://develop.redesign-aa3.pages.dev/
    * https://redesign-aa3.pages.dev/

2. Use the snippet below in the dev tools console. It prompts a change from Mainnet to Fuji and immediately afterwards reads the addresses. If everything works correctly, you should see a 100 `DEBUG ok` logs. Consistently.

```js
const switchEnv = async (isTestnet) => {
    await avalanche.request({
        id: crypto.randomUUID(),
        method: 'wallet_switchEthereumChain',
        params: [{ chainId: `0x${(isTestnet ? 43113: 43114).toString(16)}` }]
    });
}

const getAddressP = async () => {
    const [firstAcc] = await avalanche.request({ id: crypto.randomUUID(), method: 'avalanche_getAccounts' });
    
    return firstAcc.addressPVM;
}

for (let i = 0; i < 100; i++) {
    const toTestnet = i % 2 !== 1;
    await switchEnv(toTestnet);
    const addr = await getAddressP();

    if ((toTestnet && addr.startsWith('P-avax')) || (!toTestnet && addr.startsWith('P-fuji'))) {
        console.log('DEBUG failed at', i, 'switching to testnet?', toTestnet);
        break;
    } else {
        console.log('DEBUG', i, 'ok');
    }
}
```

**⚠️ Additionally, a regression of the account handling would be nice!**

## Checklist for the author


packages/common/src/concurrency/ReadWriteLock.ts
```diff
@@ -0,0 +1,87 @@
...
+  async acquireReadLock() {
+    await this.#mutex.lock();
+    try {
+      if (this.#writeCount > 0 || this.#waitingWriters > 0) {
+        const readPromise = new Promise<void>((resolve) => {
+          const handler = () => {
+            this.#emitter.removeListener('writeComplete', handler);
+            resolve();
+          };
+          this.#emitter.on('writeComplete', handler);
```
csabbee, line 29: I think you could do this:
```typescript
this.#emitter.once('writeComplete', resolve)
```
So you could get rid of defining a handler and dealing with the removal
[reference](https://nodejs.org/api/events.html#emitteronceeventname-listener)

PR #343, opened 2025-07-16
[Legacy] Solana support announcement modal

meeh0w:
* [CP-11427](https://ava-labs.atlassian.net/browse/CP-11427)

## Description
* Adds `Solana Now Supported!` modal to be shown when Solana becomes available.
* Replaces the previous prompt for deriving Solana keys for Ledger/Seedless users with the new, more generic modal.
    * We also had a modal that was shown for Seedless users informing them that keys derivation is in-progress and they can close the modal. Instead, a toast message is shown - same as Core Mobile is doing.
* I reverted the recent change that prevented users from switching to Solana when their active **_account_** did not support Solana (had no address derived). Instead, I'm only disabling the network item if the active _**wallet**_ does not support a given network (e.g. QR-based Keystone and X/P chains or Solana).
    * This allows users of Solana-supporting wallets to still switch to it and be prompted to derive keys (see videos below)

## Testing
1. Mnemonic wallets - should only be shown the modal once, right after loading up the extension.
2. Ledger wallets - should be shown the modal right after loading up the extension.
    * If users click "Get Started", they should be navigated to a full-screen tab where Ledger+Solana can be connected
    * If users close the modal, nothing happens.
    * A separate modal should be shown on each attempt to switch to Solana if the active account does not have the Solana address yet (`Add a Solana Ledger Account` instead of `Solana Now Supported!`).
3. Seedless wallets - should be shown the modal right after loading up the extension.
    * The modal that was telling users that "keys derivation is in progress" is now replaced with a toast instead for less-intrusive experience. Core Mobile does the same.
    * The aforementioned toast is only shown if keys derivation takes longer than 5 seconds.
        * If keys are derived successfully, a success message is shown
        * If an error occurs, nothing is shown. Extension will try again later (not sooner than 1 minute after the last attempt).

### With seed phrase wallets

https://github.com/user-attachments/assets/6dd112ad-dbca-4d6e-af3e-f38ff7b1649f

### With Ledger wallets

https://github.com/user-attachments/assets/cdf80d58-de4a-4e49-bd65-7f64aa257037

https://github.com/user-attachments/assets/d67a8f76-580a-4b88-a430-d3c41eb91c66

### With Seedless wallets
_timings adjusted for shorter video_

https://github.com/user-attachments/assets/e1204f9d-b3b2-4afb-9118-fec20bbbd061

### With Keystone wallets
_(or others which do not support Solana)_

https://github.com/user-attachments/assets/04f6c562-841f-4e54-94b0-d9a1be7c9a62

## Checklist for the author


packages/common/src/utils/isChainSupportedByWallet.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  switch (networkType) {
+    case NetworkVMType.EVM:
+      return true;
+    case NetworkVMType.AVM:
+    case NetworkVMType.PVM:
+    case NetworkVMType.CoreEth:
+      return (
+        secretType === SecretType.Mnemonic ||
```
csabbee, line 19: This is just a nice to have, but how about instead of writing it out like this, using some functions? 😅 
```typescript
case NetworkVMType.AVM:
case NetworkVMType.PVM:
case NetworkVMType.CoreEth:
  return [
    SecretType.Mnemonic,
    SecretType.Seedless,
    SecretType.PrivateKey,
    SecretType.Ledger,
    SecretType.LedgerLive,
    SecretType.Keystone3Pro,
  ].includes(secretType);
```
\#niceToHave

csabbee (approved):
I only have a nice to have comment, otherwise  awesome!

packages/common/src/utils/isChainSupportedByWallet.ts
meeh0w: I'll smuggle it in some future PR 🍻

PR #346, opened 2025-07-17
[Legacy] Fix missing Solana keys for mnemonic wallets

meeh0w:
* [CP-11576](https://ava-labs.atlassian.net/browse/CP-11576)
* ⚠️ Based on #343 

## Description
@rictorlome encountered an issue where some of his accounts did not have a Solana address derived (missing `addressSVM` property)

After some debugging I found the problem -- those accounts must have been created during a time when `solana-support` feature flag has been disabled, which would cause the Solana public key not to be derived upon account creation.

## Changes
With this change, extension will look for missing keys upon unlocking and derive them as needed.
We already have similar mechanisms in place for Seedless and Ledger, because it was natural those would not support Solana out-of-the-box, but the mechanism was missing for the mnemonic wallets.

## Testing
1. Run this snippet to simulate the `solana-support` feature flag being disabled:

    ```js
    await chrome.storage.local.set({ '__feature-flag-overrides__': { data: {  'solana-support': false }} })
    ```

2. Create an account. It should not have the Solana address key set.
3. Run this snippet to simulate the `solana-support` feature flag being toggled back on:

    ```js
    await chrome.storage.local.set({ '__feature-flag-overrides__': { data: {  'solana-support': true }} })
    ```

4. Lock extension
5. Unlock extension.
6. Solana address should now be derived.

### Before

https://github.com/user-attachments/assets/6a7b90bc-2c42-4867-b967-c27c3d23cbc9

### After

https://github.com/user-attachments/assets/35ce8482-00de-4916-9da0-26ab13d9c07b

## Checklist for the author

### DIFF
--- a/packages/common/src/monitoring/sentryCaptureException.ts
+++ b/packages/common/src/monitoring/sentryCaptureException.ts
@@ -41,6 +41,8 @@ export enum SentryExceptionTypes {
   FIREBASE = 'firebase',
 
   NOTIFICATIONS = 'notifications',
+
+  ACCOUNTS = 'accounts',
 }
 
 // wrapper to make error reporting contexts unfirom accross the codebase
--- a/packages/common/src/utils/account.ts
+++ b/packages/common/src/utils/account.ts
@@ -1,6 +1,22 @@
 import { Account } from '@core/types';
 
 export function getAllAddressesForAccount(acc: Partial<Account>) {
+  return getAllAddressesForAccountUnfiltered(acc).filter(
+    (addr): addr is string => typeof addr === 'string',
+  );
+}
+
+export function getAllAddressesForAccounts(accounts: Account[]): string[] {
+  return accounts
+    .flatMap(getAllAddressesForAccount)
+    .filter((v) => typeof v === 'string');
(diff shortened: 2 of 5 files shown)


apps/legacy/src/components/common/header/NetworkSwitcher/NetworkSwitcher.tsx
```diff
@@ -143,38 +142,61 @@ export function NetworkSwitcher() {
...
-                            sx={{ alignItems: 'center', flexDirection: 'row' }}
+                          <NetworkSelectronMenuItem
+                            data-testid={`select-network-${networkItem.chainId}-button`}
+                            onClick={() => {
+                              if (!isSupported) {
+                                return;
+                              }
+                              setNetwork(networkItem);
+                              setIsOpen(false);
+                            }}
```
bartosz-the-coder-at-avalabs, line 169:
```suggestion
                            onClick={isSupported 
                            ?() => {
                              setNetwork(networkItem);
                              setIsOpen(false);
                            }
                            : undefined}
```

PR #351, opened 2025-07-22
Cp 11431 keystore

frichards:
## Checklist for the author


meeh0w (changes requested):
Left some comments :)

Please try not to override Mui classes for components that should just work -- instead use the closest variant and leave a `TODO` comment to revisit after typography improvements in K2 is finished.

@bartosz-the-coder-at-avalabs also taught me that `sx` is less performant than using props directly (e.g. `<Stack gap={2}>` performs (and, imho, reads) better than `<Stack sx={{ gap: 2 }}>`).

apps/next/src/pages/AccountManagement/components/ImportKeystoreFile/Page.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const reset = useCallback(() => {
+    setError(null);
+    setFile(null);
+    setStep(Step.ChooseFile);
+  }, []);
+
+  const onFileSelected = useCallback(
+    async (rawFile: File | null) => {
+      if (!rawFile) {
+        setError(KeystoreError.InvalidVersion);
```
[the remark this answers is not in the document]
frichards: I also had to double check when I copied the error from logacy. 
https://github.com/ava-labs/core-extension/blob/main/apps/legacy/src/pages/Accounts/AddWalletWithKeystoreFile.tsx#L77
I am good to set another error here if we want though. 
The current KeystoreError does not include suitable error I do not think. And unknown error is too generic here. So we would have to add an error, I think. What do you think?

apps/next/src/pages/AccountManagement/components/ImportKeystoreFile/components/KeystoreFilePassword.tsx
```diff
@@ -15,25 +27,38 @@ export const KeystoreFilePassword = ({
...
+      capture('KeystoreFileImportStarted');
+    },
+    onSuccess: () => {
+      capture('KeystoreFileImportSuccess');
+      toast.success(t('Successfully imported the keystore file.'));
+      replace('/account-management');
+    },
+  };
+
+  const { importFile, isImporting } = useImportKeystoreFile(importCallbacks);
```
meeh0w, line 53: Let's wrap it in `useMemo`, otherwise the callbacks will be redefined on every state change, making any hooks and callbacks inside `useImportKeystoreFile` to get redefined too :)

apps/next/src/pages/AccountManagement/components/ImportKeystoreFile/hooks/useImportKeystoreFile.ts
```diff
@@ -1,38 +1,38 @@
...
-        onError(err);
+        onSuccess?.();
+      } catch {
+        onFailure?.();
       } finally {
         setIsImporting(false);
       }
     },
-    [isImporting, capture, importKeystoreFile, replace, t],
+    [isImporting, importKeystoreFile, onStarted, onSuccess, onFailure],
```
meeh0w: If we add the callbacks to `useKeystoreFileImport()` hook (the one being used in line `15` here), this new hook (`useImportKeystoreFile`) will no longer be needed.

Unless I'm misunderstanding something :)

apps/next/src/pages/AccountManagement/components/ImportKeystoreFile/Page.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { KeystoreFileUpload } from './components/KeystoreFileUpload';
+import { KeystoreFilePassword } from './components/KeystoreFilePassword';
+import { KeystoreFileError } from './components/KeystoreFileError';
+
+enum Step {
+  ChooseFile,
+  ProvidePassword,
+  ConfirmData,
+  Error,
+}
```
bartosz-the-coder-at-avalabs, line 14: Please consider using string union.
It's much more convenient.

bartosz-the-coder-at-avalabs (dismissed):

meeh0w (dismissed):
✨

meeh0w (approved):

apps/next/src/pages/AccountManagement/components/ImportKeystoreFile/components/FileImage.tsx
```diff
@@ -0,0 +1,19 @@
...
+const Dark = () => {
+  return <img src={DarkImage} alt="File Image" width={44} height={51} />;
+};
+
+export const FileImage = () => {
+  const theme = useTheme();
+
+  const image = theme.palette.mode === 'light' ? <Light /> : <Dark />;
+
+  return image;
```
bartosz-the-coder-at-avalabs, line 18: The file can be simplified even more: 
```suggestion
export const FileImage = () => {
  const theme = useTheme();

  const image = theme.palette.mode === 'light' ? LightImage : DarkImage

  return <img src={Image} alt="File Image" width={44} height={51} />;
```

This way you don't need to remember to sync width and height of multiple components

apps/next/src/pages/AccountManagement/components/ImportKeystoreFile/components/KeystoreFilePassword.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  return (
+    <Stack sx={{ flexGrow: 1, alignItems: 'center' }}>
+      <Card sx={{ p: '12px' }}>
+        <Stack
+          sx={{
+            flexDirection: 'row',
+            justifyContent: 'space-between',
+            alignItems: 'center',
+            columnGap: '11px',
```
bartosz-the-coder-at-avalabs, line 76: Please use design system responsive values instead of the fixed ones
```suggestion
        <Stack
          sx={{
            flexDirection: 'row',
            justifyContent: 'space-between',
            alignItems: 'center',
            columnGap: theme => theme.spacing(1.25)
          }}
```

bartosz-the-coder-at-avalabs: also, for the container components (Box, Stack & Grid) there's no need to use `sx` prop, because they export most of css properties as own props.
The notation is more concise and impacts the memory less  

``` tsx
<Stack
  direction="row"
  justifyContent='space-between"
  alignItems="center"
  columnGap={1.25}
>
```

bartosz-the-coder-at-avalabs (changes requested):

PR #373, opened 2025-08-08
fix: CP-11379 pages, typographies, forgotpassword etc

vvava:
## Description

There are multiple issues in the current implementation of the new design.

## Changes

Fixing some of them but there are still work to do e.g. [the account manager page](https://ava-labs.atlassian.net/jira/software/c/projects/CP/boards/72/backlog?epics=visible&issueParent=65763&selectedIssue=CP-11801)

## Testing

Go to the extension and click through the pages and compare them to the designs -> it should be checked in the sidebar mode as well.

## Checklist for the author


apps/next/src/components/SwitchCard/SwitchCard.tsx
```diff
@@ -3,40 +3,56 @@ import {
...
+      variant: 'subtitle3' as TypographyVariant,
+      // TODO: fix it after a proper k2 release
+      fontWeight: 600,
+    },
+    large: {
+      variant: 'h5' as TypographyVariant,
+      fontWeight: theme.typography.fontWeightBold,
+    },
+  };
+  console.log('theme: ', theme);
```
meeh0w: ✂️

apps/next/src/components/SwitchCard/SwitchCard.tsx
```diff
@@ -3,40 +3,56 @@ import {
...
   return (
     <StyledSwitchCard orientation={orientation}>
-      <Stack gap={1}>
-        <Typography {...titleProps}>{title}</Typography>
+      <Stack gap={0.5}>
+        <Typography
+          variant={titleStyles[titleSize].variant}
+          component="h3"
+          fontWeight={titleStyles[titleSize].fontWeight}
```
meeh0w, line 51: Why not `{...titleStyles[titleSize]}` instead of listing all props by hand?

apps/next/src/components/SwitchCard/SwitchCard.tsx
```diff
@@ -3,40 +3,56 @@ import {
...
+    small: {
+      variant: 'subtitle3' as TypographyVariant,
+      // TODO: fix it after a proper k2 release
+      fontWeight: 600,
+    },
+    large: {
+      variant: 'h5' as TypographyVariant,
+      fontWeight: theme.typography.fontWeightBold,
+    },
+  };
```
meeh0w, line 44: Since this is constant, can it be moved outside of the component body? Instead of `theme.typograpy.fontWeightBold`, just `fontWeightBold` should work.

apps/next/src/components/SwitchCard/SwitchCard.tsx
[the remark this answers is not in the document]
vvava: as usual we can say :D

apps/next/src/pages/Settings/components/Home/components/SettingsCard.tsx
```diff
@@ -13,11 +14,11 @@ export const SettingsCard: FC<SettingsCardProps> = ({
   description,
   ...cardProps
 }) => {
+  const theme = useTheme();
   return (
     <StyledCard {...cardProps}>
       <Stack gap={0.5}>
-        {/* TODO: find a common ground with UX team so we don't have to override the fontWeight*/}
-        <Typography variant="h5" fontWeight="700 !important">
+        <Typography variant="h5" fontWeight={theme.typography.fontWeightBold}>
```
meeh0w, line 21: You should be able to use `fontWeight="fontWeightBold"`, no need to `useTheme()`.

frichards (dismissed):
🌈 🦄 Approved!

meeh0w (changes requested):
Thanks for taking care of those discrepancies! I noticed a few things, listed below.

apps/next/src/pages/AccountManagement/AccountManagement.tsx
```diff
@@ -42,67 +38,50 @@ const dialogSlots: Pick<DialogProps, 'slots' | 'slotProps'> = {
   },
 };
 
-const dialogContentSx: SxProps = {
+export const dialogContentSx: SxProps = {
```
meeh0w, line 41: If you need to use those things outside of `AccountManagement`, let's either move it out of this file, or just duplicate them (nothing wrong with that) -- especially since `dialogContentSx` references a container specific to account-management.

The way it is right now, you need to import `pages/AccountManagement/AccountManagement` from the lock screen, just for those props.

apps/next/src/pages/LockScreen/components/ForgotPassword.tsx
```diff
@@ -16,7 +16,7 @@ import { ResetExtensionStateHandler } from '~/services/storage/handlers/resetExt
 
 import { WarningMessage } from '@/components/WarningMessage';
 import { PageTopBar } from '@/components/PageTopBar';
-import { dialogSlots } from '@/pages/AccountManagement';
+import { dialogSlots } from '@/pages/AccountManagement/components/Dialog';
```
meeh0w, line 19: It's still cross-referencing to `AccountManagement` from a different page. It needs to go higher.

Since we want `<Dialog />` to have the same default `slots` and `slotProps`, we should create `@/components/Dialog` that pre-defines them, and use this new `<Dialog />` with default values in both `AccountManagement` and `in `ForgotPassword`.

Oor, place those defaults props in `@/config/dialog` and reference it from both places.

meeh0w (changes requested):

meeh0w (approved):
Thanks for the updates! 🎉

PR #374, opened 2025-08-08
[Next Gen] Settings > Connected sites

bartosz-the-coder-at-avalabs:
## Description

https://ava-labs.atlassian.net/browse/CP-11527
https://www.figma.com/design/kksuPP4PnRIFDx9Sm3544F/Core-Extension-Redesign-2025?node-id=104-954&m=dev

## Screenshots:

https://github.com/user-attachments/assets/f0f73fa2-2dad-482c-9ca5-c62ef773c9c1

## Checklist for the author


apps/next/src/pages/Settings/components/ConnectedSites/components/ConnectedSiteItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    isFaviconExists(site.domain).then(setIsIconExists);
+  }, [site.domain]);
+
+  return (
+    <Styled.ListItem>
+      <Styled.Avatar
+        src={isIconExists ? site.favicon : undefined}
+        alt={displayName}
+        error={!isIconExists}
```
meeh0w, line 32: Do we have to check existence this way? How about we instead just try to fetch the image and fall back to some default with an `onError` event handler?

bartosz-the-coder-at-avalabs: If we want to comply the Figma then yes.
Without the check, HTTP 404 returns a placeholder image
image

It looks ok, but totally different than the placeholder on Figma.

I tried to listen on `error` event of Avatar and underlaying img element, but with no success :(

meeh0w:
From what I can tell, the `AccountSelector` is pretty close to the designs -- maybe some typography is different in the group headers, but no biggie.

If it needs a change, please point out which parts exactly @vvava 🙇  and I'll fix it in another PR -- fixing it here will make it conflict with my upcoming PR.

csabbee (approved):
Tested, working as described

PR #376, opened 2025-08-12
feat(next-gen): implement TokenSelect and TokenAmountInput

meeh0w:
* [CP-11822](https://ava-labs.atlassian.net/browse/CP-11822)

## Changes
Implements the following:
* `useTokensForAccount()` hook - which accepts an `Account`, looks at the balances for this account and returns a list of fungible tokens (non-NFTs) that the provided account holds (sorted accordingly)
* `TokenSelect` - a dropdown that accepts a token list (preferably obtained with `useTokensForAccount()` and allows querying
* `InvisibleAmountInput` - an input that has no boundaries and resizes the font-size so that the provided amount fits without overflow (until a minimum font-size is reached, at which point it allows overflowing)
* `OverflowingTypography` - a `<Typography />` wrapped in a tooltip. Tooltip is only shown when the typography element overflows (within this PR, this is used for token names -- some would not fit in the constrained space we work with)
* `TokenAmountInput` -- groups `TokenSelect` and `InvisibleAmountInput` into a nice container that allows selecting both the token and it's amount (also renders buttons for predefined amounts - `25%` / `50%` / `Max`)

## Screenshots:

https://github.com/user-attachments/assets/d6057834-2171-4718-905d-a4cf382df6b4

## Checklist for the author


csabbee (dismissed):
just one minor issue found, regarding the search field not getting cleared after selecting a token, but that will be addressed in a followup PR

apps/next/src/components/TokenAmountInput/components/InvisibleAmountInput/hooks/useDynamicFontSize.tsx
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+
+    setFontSize(suggestedFontSize);
+  }, [element, text, maxWidth, minFontSize, maxFontSize]);
+
+  const measureElement = useCallback(
+    () => <MeasureElement ref={element}>{text}</MeasureElement>,
+    [text],
+  );
```
bartosz-the-coder-at-avalabs, line 46: To be exact it's not an element, but a component 😛 
It should be called using triangular brackets rather than parentheses.

bartosz-the-coder-at-avalabs (dismissed):

apps/next/src/components/TokenAmountInput/components/InvisibleAmountInput/hooks/useDynamicFontSize.tsx
meeh0w: true :D

PR #392, opened 2025-08-19
fix(onboarding): reset password

frichards:
## Description
Ticket: https://ava-labs.atlassian.net/browse/CP-11885

When the user navigate back to any previous step in onboarding after filling out the wallet details page and get back to the wallet details page, the original password should not be stored. 

## Changes
I created resetWalletDetails function in the onboarding provider. 
And it gets called every time the wallet details page gets mounted to ensure that the form has default values. 

## Testing
Please go through the onboarding process and fill out the wallet details. 
In stead of next, click back and get back to wallet details. Any info you filled out in wallet details should be gone. 

## Screenshots:

https://github.com/user-attachments/assets/f19f7a83-1a63-4afb-8d18-1deda416290f

## Checklist for the author


meeh0w (changes requested):
I don't really like this solution, because if I click `Next` on the wallet details screen (to land on the `Customize Core to your liking` screen) and then change my mind about any of the fields (i.e. want to use a different name for the wallet or want to subscribe to the newsletter after all), I have to populate all of the fields from scratch.

I think what @vvava meant in the ticket is that those details should be cleared when the user goes back to the very first screen of the onboarding flow, meaning this:

Image

apps/next/src/pages/Onboarding/common-screens/ProvideWalletDetailsScreen/ProvideWalletDetailsScreen.tsx
```diff
@@ -37,6 +37,14 @@ export const ProvideWalletDetailsScreen: FC<
   const { t } = useTranslation();
   const { setCurrent, setTotal } = useModalPageControl();
 
+  const { resetWalletDetails } = useOnboardingContext();
+
+  // Reset Wallet Details fields each time this screen is mounted
+  useEffect(() => {
+    resetWalletDetails();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
bartosz-the-coder-at-avalabs, line 46: That's kind of hacky.
Since `resetWalletDetails` function is already wrapped with `useCallback` it will not cause any extra re-renders.
In its current state, the code will be difficult to extend because each attempt will require checking why the ESLint directive is present instead of the dependency being added to the deps array.
And the answer here is just: there's no reason

apps/next/src/pages/Onboarding/Onboarding.tsx
```diff
@@ -39,6 +40,20 @@ export function Onboarding() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  useEffect(() => {
+    const maybeReset = () => {
+      console.log(window.location.hash);
```
bartosz-the-coder-at-avalabs, line 45: Console Log 😜

apps/next/src/pages/Onboarding/Onboarding.tsx
```diff
@@ -39,6 +40,20 @@ export function Onboarding() {
...
+      if (window.location.hash === '#/onboarding') {
+        resetStates();
+      }
+    };
+
+    window.addEventListener('hashchange', maybeReset);
+    // Run on mount as well
+    maybeReset();
+    return () => window.removeEventListener('hashchange', maybeReset);
+  }, [resetStates]);
```
meeh0w, line 55: 1. Let's ✂️ the `console.log`
2. From my testing, I usually get to `home.html#/` 🤔 and only go to `home.html#/onboarding` in case of some error state. We should make sure to call `resetStates()` in both scenarios (unifying this to one of the two would also work, but I don't know if it's worth the effort _right now_ - ideally, the whole onboarding should be a separate bundle in my opinion (I have a task to make it this way somewhere)).
3. You can use `const match = useRouteMatch({ path: '/(onboarding)?/' });`. This way this won't break if we ever change from [`HashRouter`](https://v5.reactrouter.com/web/api/HashRouter) to something else. You also won't need to register event listeners, just a good old `useEffect` with `match?.isExact` as the dependency should work:

```ts
  const match = useRouteMatch({ path: '/(onboarding)?/' });

  // #/                   ===> match.isExact: true
  // #/onboarding         ===> match.isExact: true
  // #/onboarding/import  ===> match.isExact: false
  // etc...

  useEffect(() => {
    if (match?.isExact) {
      resetStates();
    }
  }, [resetStates, match?.isExact]);
```

bartosz-the-coder-at-avalabs: I'd rather stay with Fumiko's hashchange event listener approach,
but instead of this:
``` ts
if (window.location.hash === '#/onboarding') {
        resetStates();
      }
```
we can use `matchRoute` function instead 
``` ts
const [, hash] = e.newURL.split('#');
if (matchPath(hash, { path: "onboarding?", exact: true }) {
        resetStates();
    }
}
```

meeh0w (approved):

bartosz-the-coder-at-avalabs (approved):

PR #394, opened 2025-08-20
refactor: move getMaxUtxoSet() out to shared utils

meeh0w:
This just moves a shared util to `@core/common` namespace so it does not clutter the upcoming PRs.


frichards (approved):
🌈 🦄 Approved!

csabbee (approved):

PR #412, opened 2025-09-01
feat(ng): bitcoin approval screen

meeh0w:
* ⚠️ _Based on #411_
* [CP-11848](https://ava-labs.atlassian.net/browse/CP-11848)

## Description
* Adds the approval screen for Bitcoin transactions

## Testing
1. Perform a simple send in the app
2. Bridge some BTC from Bitcoin to Avalanche (using https://core.app/bridge)

### Send

https://github.com/user-attachments/assets/42100a77-152d-4876-8e52-6ade3f2c8bd5

### Bridge via Core.app

https://github.com/user-attachments/assets/6d6878a9-2ea0-4112-b8e1-e8b2c28c32c1

## Checklist for the author

### DIFF
--- a/apps/next/src/config/constants.ts
+++ b/apps/next/src/config/constants.ts
@@ -1,3 +1,5 @@
+import { TransactionPriority } from '@core/types';
+
 export const CORE_WEB_BASE_URL =
   process.env.CORE_WEB_BASE_URL ?? 'https://core.app';
 export const BUG_BOUNTIES_URL =
@@ -9,3 +11,6 @@ export const CORE_FEEDBACK_URL =
   'https://docs.google.com/forms/d/e/1FAIpQLSdUQiVnJoqQ1g_6XTREpkSB5vxKKK8ba5DRjhzQf1XVeET8Rw/viewform?usp=pp_url&entry.2070152111=Core%20browser%20extension&entry.903657115=${extensionVersion}&entry.1148340936=${os}';
 
 export const DARK_THEME_SURFACE_COLOR = '#404046';
+
+export const DEFAULT_FEE_PRESET: TransactionPriority = 'low';
+export const DEFAULT_FEE_PRESET_C_CHAIN: TransactionPriority = 'high';
--- a/apps/next/src/pages/Approve/components/ActionDetails/btc/components/BtcNetworkFeeWidget/hooks/useBtcTransactionFee/useBtcTransactionFee.tsx
+++ b/apps/next/src/pages/Approve/components/ActionDetails/btc/components/BtcNetworkFeeWidget/hooks/useBtcTransactionFee/useBtcTransactionFee.tsx
@@ -5,6 +5,7 @@ import { useConnectionContext } from '@core/ui';
 import { calculateGasAndFees } from '@core/common';
 import { type UpdateActionTxDataHandler } from '@core/service-worker';
 
+import { DEFAULT_FEE_PRESET } from '@/config';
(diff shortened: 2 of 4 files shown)


apps/next/src/pages/Approve/components/ActionDetails/btc/components/BtcNetworkFeeWidget/types.ts
```diff
@@ -0,0 +1 @@
+export type BtcFeePreset = 'slow' | 'normal' | 'fast';
```
gergelylovas, line 1: Question. Why are we introducing a mapping on the type level for the values from the bitcoin vm-module? Should this be simply a mapping on the button text level? Feels like it would require less mental strain to follow what slow normal and fast means

apps/next/src/pages/Approve/components/ActionDetails/btc/lib/btcDetailsUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+type MapFunction = (
+  network: BtcNetwork,
+  t: TFunction,
+) => (section: DetailSection, index: number) => DetailSection;
+
+// The funds recipients section for NextGen looks different than from the legacy one.
+// To keep things backwards-compatible, we map the recipients items here instead of
+// updating the Bitcoin Module.
+// FIXME: This should be changed (in sync with Core Mobile) once we fully migrate
```
gergelylovas, line 14: Please create a tech debt ticket for this.

gergelylovas (changes requested):
Question: In your recordings the approval screen for in-extension send and dapp approval look different beyond the fee section. 
Is there any particular reason for the difference? I feel having the same values in the same order on the approval screen is a safer and more consistent approach.

meeh0w:
@gergelylovas The "Send" operation triggers a `bitcoin_sendTransaction`, while the "Bridge" operation triggers a `bitcoin_signTransaction` (since updating the fees on Bitcoin means reconstructing the whole transaction from scratch, the Unified Bridge SDK would no longer be the source of the transaction - that's why it just supplies the tx to be signed, but not customized).

So long story short, the differences in the UI come from the implementations in the Bitcoin Module (which, to be fair, could've been more consistent).

Ideally though, both request types could use Blockaid to simulate the transaction and then we could display the proper balance changes (tx simulation was not a thing back when either of the requests were implemented in the Bitcoin module).

apps/next/src/pages/Approve/components/ActionDetails/btc/components/BtcNetworkFeeWidget/types.ts
meeh0w: Yup, renamed those to match the keys in `NetworkFees` object.

apps/next/src/pages/Approve/components/ActionDetails/btc/lib/btcDetailsUtils.ts
meeh0w: Yup -> [CP-11995](https://ava-labs.atlassian.net/browse/CP-11995)

gergelylovas (approved):

gergelylovas:
> So long story short, the differences in the UI come from the implementations in the Bitcoin Module (which, to be fair, could've been more consistent).
@meeh0w makes sense, can you please create a ticket for this as well. Even if we don't enable TX simulation for BTC we should make sure that the approval windows are consistent.

PR #416, opened 2025-09-03
feat(network): add custom network flow

frichards:
## Description
Ticket: https://ava-labs.atlassian.net/browse/CP-11700
Part 3

Based on this PR: https://github.com/ava-labs/core-extension/pull/415

## Changes
Now a user can add a custom network
Because of the change of the flow, when we store the custom network for the first time, we allow the custom rpc headers to be stored at the same time.

## Testing
Go to network management, click on plus sign at the top right. 
Fill out the form and save. 
The new network should be saved and under custom network list, you should see the newly added network. 

## Screenshots:
Screenshot 2025-09-03 at 15 34 59

## Checklist for the author

- [X] I've covered new/modified business logic with Jest test cases.


apps/next/src/components/Forms/FormField.tsx
```diff
@@ -98,12 +102,14 @@ export const FormField = ({
           >
             <Stack component="label" onClick={() => setIsEditing(true)}>
               <InvisibleFieldInput
+                type={type}
```
meeh0w, line 105: Can we make it such that the error message is only shown if the field was touched/focused?
This way we won't be displaying this scary all-red form as soon as the user enters the page.

Quick idea:

```tsx
  const [wasTouched, setWasTouched] = useState(false);

  const showPrompt = Boolean(value || isEditing || (wasTouched && required));

  const handleChange = useCallback(
    (e: React.ChangeEvent<HTMLInputElement>) => {
      setWasTouched(true);
      onChange(e.target.value);
    },
    [onChange],
  );

  // then use `handleChange` on the input field
  // and only render `error` / change color to red if `wasTouched` is true
```

apps/next/src/pages/Settings/components/NetworkManagement/hooks/useAddNetworks.ts
```diff
@@ -0,0 +1,76 @@
...
+export const useAddNetwork = () => {
+  const { saveCustomNetwork } = useNetworkContext();
+  const { capture } = useAnalyticsContext();
+
+  const [network, setNetwork] = useState<Network & AdvancedNetworkConfig>(
+    defaultNetworkValues,
+  );
+  const { isValid, errors } = useMemo(() => {
+    return isNetworkValid(network);
+  }, [network]);
```
meeh0w, line 35: Something that Bartosz made me alert to -- `useMemo` is not needed for things like this.

Whenever you update any field, `network` will change and therefore `useMemo` will have to run again, giving us no benefit, but we'll still be paying the cost of memoization.

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -665,7 +665,7 @@ export class NetworkService implements OnLock, OnStorageReady {
 
     this._customNetworks = {
       ...this._customNetworks,
-      [chainId]: omit(customNetwork, 'customRpcHeaders') as NetworkWithCaipId, // should be saved in overrides
```
meeh0w, line 668: Is this an intended change?

meeh0w (changes requested):
Looks great, I left a few smol comments :)

PR #432, opened 2025-09-16
feat: CP-11851 Core concierge

vvava:
## Description

[Implement the Concierge with the new UI.](https://ava-labs.atlassian.net/jira/software/c/projects/CP/boards/72?selectedIssue=CP-11851)

## Changes

Put the button and the animation to the header in the portfolio page. Create the Concierge page.

## Testing

Go to Concierge page and play around with it. Also try the turn on and off in settings and feature flag out.

## Screenshots:

https://github.com/user-attachments/assets/8647ecef-2da3-4065-9dff-be8fb906a48e

https://github.com/user-attachments/assets/6bf3bce7-a240-4fe4-88b9-53a15bac4b47

## Checklist for the author


vvava:
Feature flags are missing

apps/next/src/pages/Settings/components/Home/Home.tsx
```diff
@@ -158,16 +164,18 @@ export const SettingsHomePage = () => {
...
+      {featureFlags[FeatureGates.CORE_ASSISTANT] && (
+        <SwitchCard
+          title={t('Core Concierge')}
+          titleSize="large"
+          orientation="horizontal"
+          description={t(
+            'Get Core to work for you. Whether it’s transferring, sending crypto, just ask away!',
+          )}
+          checked={coreAssistant}
+          onChange={() => setCoreAssistant(!coreAssistant)}
```
bartosz-the-coder-at-avalabs, line 176:
```suggestion
          onChange={() => setCoreAssistant(is => !is)}
```

apps/next/src/components/Page/Page.tsx
```diff
@@ -54,7 +56,7 @@ export const Page = ({
         title={title}
       />
       <NoScrollStack>
-        <Stack px={1.5} pb={1.5} gap={3} flexGrow={1} {...containerProps}>
+        <Stack px={py ?? 1.5} pb={1.5} gap={3} flexGrow={1} {...containerProps}>
```
bartosz-the-coder-at-avalabs, line 59: "px={py}"?
Is this correct?

apps/next/src/pages/Concierge/components/UserInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+          if (message) {
+            setInput(message);
+          }
+          if (index > 0) {
+            setUserMessageHistoryIndex(index);
+          }
+        }
+        if (e.key === 'ArrowDown' && userMessages) {
+          const index =
+            userMessageHistoryIndex || userMessageHistoryIndex === 0
```
bartosz-the-coder-at-avalabs, line 97:
```suggestion
            userMessageHistoryIndex >= 0
```

apps/next/src/pages/Concierge/components/UserInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+        if (e.key === 'Enter' && input) {
+          e.preventDefault();
+          setPrompt(input);
+          setUserMessageHistoryIndex(userMessages?.length || 1);
+        }
+      }}
+      onKeyUp={(e) => {
+        if (e.key === 'ArrowUp' && userMessages) {
+          const index =
+            userMessageHistoryIndex || userMessageHistoryIndex === 0
```
bartosz-the-coder-at-avalabs, line 83:
```suggestion
            userMessageHistoryIndex  >= 0
```

apps/next/src/pages/Concierge/components/UserInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+          if (index > 0) {
+            setUserMessageHistoryIndex(index);
+          }
+        }
+        if (e.key === 'ArrowDown' && userMessages) {
+          const index =
+            userMessageHistoryIndex || userMessageHistoryIndex === 0
+              ? userMessageHistoryIndex + 1
+              : userMessages.length - 1;
+          const message = userMessages[index];
```
bartosz-the-coder-at-avalabs, line 100: There's a bug.
`userMessageHistoryIndex + 1` can result in a value greater than `userMessages.length`

bartosz-the-coder-at-avalabs (changes requested):

apps/next/src/pages/Settings/components/Home/Home.tsx
vvava: Why is this approach better? It seems to have an additional function call.

apps/next/src/pages/Concierge/components/AIDialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+            overflow: 'hidden',
+            wordBreak: 'break-word',
+          }}
+        >
+          {lastMessage && !isTextTyped && (
+            <Typography>
+              <Typewriter
+                text={message.content}
+                scrollToBottom={scrollToBottom}
+                typingSpeed={typingSpeed}
```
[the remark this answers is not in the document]
vvava: So the reason why it lives there because the Typewriter should not care about the message length it only needs just the speed and I don't want to the TypeWriter contains extra "business logic" it should be only a "presentation" component. Therefor the component can be reusable without any issues in the typing speed (e.g. we will need the use the component but we don't want to calculate the speed based on the message length).

apps/next/src/pages/Concierge/hooks/useFunctions.ts
```diff
@@ -0,0 +1,691 @@
...
+      switchNetwork: async ({ networkId }: { networkId: string }) => {
+        const newActiveNetwork = getNetwork(networkId);
+        if (!newActiveNetwork) {
+          throw new Error(`Cannot find the new network you want to activate.`);
+        }
+        setNetwork(newActiveNetwork);
+        return {
+          content: `Success! The new active network is ${newActiveNetwork.chainName}`,
+        };
+      },
```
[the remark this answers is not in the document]
meeh0w: That's a good idea 👌

apps/next/src/pages/Concierge/components/AIDialog.tsx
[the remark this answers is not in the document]
meeh0w: But the text length is clearly the single thing it actually cares about when it comes to the typing speed. It's not business logic, it's the presentation logic.

And tbh I think deriving the typing speed based on the text length is the only thing that makes sense, otherwise the component would become super clunky imho (I mean, it's good that longer text translates into higher typing speed).

The only other way I could see it be is that the message typing duration could always be constant (e.g. no matter the content, it would always take 1s for the message to appear). But I feel like this could also not be super user-friendly.

Just saying, leave it be if you wish -- no action required.

apps/next/src/pages/Settings/components/Home/Home.tsx
meeh0w: You should use a state setter's callback when the new state depends on the previous state, such as incrementing a counter or reversing boolean flags. Using the callback ensures your update is based on the most current state, preventing issues from React's state update batching and potential race conditions.

Low risk here, but it's a good thing to remember especially with more complex state variables (e.g. objects, arrays). Also useful inside `useEffect` or `useCallback` - you don't need the state itself as a dependency.

vvava: I understand that is the case when we use `useState()` but here the setCoreAssistant is a network request (and it comes from a context) so we must set the opposite of the latest value which is the correct value there. In addition since it is not a React state setter  it won't accept this form.

meeh0w: Oohhh I see, I think both me and Bartosz got confused by the `set..()` naming here. My bad.

frichards (approved):
🌈 🦄 Approved!
Manually tested

apps/next/src/components/Header/ConciergePrompt.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import {
+  AnimatedButton,
+  PromptButtonBackground,
+  TextAnimation,
+} from './components/styledComponents';
+import { useFeatureFlagContext, useSettingsContext } from '@core/ui';
+import { FeatureGates } from '@core/types';
+
+export const ConciergePrompt = ({ isAIBackdropOpen, setIsAIBackdropOpen }) => {
```
bartosz-the-coder-at-avalabs: Missing props type

apps/next/src/components/Header/ConciergePrompt.tsx
```diff
@@ -0,0 +1,218 @@
...
+                },
+                '.overlay-exit.prompt-background': {
+                  display: 'block',
+                  opacity: 1,
+                },
+                '.overlay-exit-done.prompt-background': {
+                  display: 'block',
+                  opacity: 0,
+                },
+              }}
```
bartosz-the-coder-at-avalabs: Please consider moving all styled components to a separate file(s).
The styles overrides are quite big and overshadow the actual logic, making it harder to understand.

bartosz-the-coder-at-avalabs (changes requested):

apps/next/src/components/Header/components/styledComponents.ts
```diff
@@ -0,0 +1,157 @@
...
+      opacity: '0',
+      transform: 'scale(0)',
+    },
+  },
+  '&.button-enter': {
+    span: {
+      display: 'inline',
+    },
+  },
+  '&.button-enter-done': {
```
bartosz-the-coder-at-avalabs: I guess you missed these two spots 😞

PR #441, opened 2025-09-24
fix: remove platformFee when no feeAccount

gergelylovas:
## Description

Fixes issue when swapping SLP tokens we don't have a fee account initialized for.

## Changes

Removes platformFee property from the quote for tokens without fee accounts.

## Testing

Do a solana swap with a destination token being something less popular.

## Checklist for the author


frichards (approved):
🌈 🦄 Approved!

PR #443, opened 2025-09-25
fix(deps): version update for sdks and fix recent recipients

frichards:
## Description

Issue: Current main has issue with ledger signing
Bonus issue: Contract address was still showing for recent recipients

## Changes
1. updated the avalanche-sdks versions. 
2. added a filter to only get the receipients from send and nft send

## Testing
Please try using ledger to send on x and p chain. 
Please check that addresses showing in recents in address book is actual address and not contracts.

## Checklist for the author


gergelylovas (approved):

vvava (changes requested):
It won't work on my end: 

https://github.com/user-attachments/assets/a4c6e67a-25ec-44c8-a8eb-8d91c8054ef0

PR #449, opened 2025-10-02
fix(pchain): make pchain wallet balance match

frichards:
## Description
https://ava-labs.atlassian.net/browse/CP-12215
https://ava-labs.atlassian.net/browse/CP-12209

## Changes
1. Link to core web for pchain balance details got updated. (This ticket is a blocker for this. https://ava-labs.atlassian.net/browse/CP-12175. But the current link is also broken on prod. )
2. Following the new pchain wallet balance strategy
3. updating AvalancheGetAddressesInRangeHandler to retutn the derived accounts addresses when the user is using ledger live or seedless 

## Testing
1. If you have any account in a wallet that is underived with pchain balance, it should show the hyper text to take you to core-web pchain wallet details. 
2. When you connect the seedless or ledger live, all the derived account xp addresses in that wallet should be return when calling "avalanche_getAddressesInRange" method. 

## Checklist for the author

### DIFF
--- a/packages/service-worker/src/services/balances/handlers/getTotalBalanceForWallet/getTotalBalanceForWallet.ts
+++ b/packages/service-worker/src/services/balances/handlers/getTotalBalanceForWallet/getTotalBalanceForWallet.ts
@@ -95,7 +95,8 @@ export class GetTotalBalanceForWalletHandler implements HandlerType {
   handle: HandlerType['handle'] = async ({ request }) => {
     const { walletId } = request.params;
     const requestsImportedAccounts = isImportedAccountsRequest(walletId);
-
+    const priceChangesData =
+      await this.balanceAggregatorService.getPriceChangesData();
     try {
       const allAccounts = await this.accountsService.getAccounts();
       const derivedAccounts = requestsImportedAccounts
@@ -134,6 +135,7 @@ export class GetTotalBalanceForWalletHandler implements HandlerType {
         derivedAddressesBalances,
         derivedAccounts,
         networksIncludedInTotal,
+        priceChangesData,
       );
       let hasBalanceOnUnderivedAccounts = false;
 
--- a/apps/legacy/src/pages/Accounts/components/WalletContainer.tsx
+++ b/apps/legacy/src/pages/Accounts/components/WalletContainer.tsx
@@ -74,7 +74,7 @@ export const WalletContainer = ({
               variant="text"
               onClick={() => {
                 window.open(
-                  `https://core.app/portfolio/p-chain/asset`,
+                  `https://core.app/portfolio/wallet/p-chain/utxos`,
                   '_blank',
                   'noreferrer',
                 );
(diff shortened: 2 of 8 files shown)


packages/service-worker/src/services/accounts/handlers/avalanche_getAddressesInRange.ts
gergelylovas, line 1: Please update the tests of this handler.

packages/service-worker/src/services/balances/handlers/getTotalBalanceForWallet/helpers/processGlacierAddresses.ts
gergelylovas, line 1: please update the tests for this file

gergelylovas (changes requested):
I think there are two differences compared to the spec at the moment:
- The AVAX price for P/C/X chain should come from the watchlist
- Balances in the atomic memory should be reflected on the source chain.

rictorlome:
@frichards -- does signing work across the accounts for this implementation? it is not enough to just show a balance rolled up across the wallet for XP chain if the user cannot sign for all those UTXOS. it is a broken experience to return account 1 + 2 from getaddressesinrange and then a user attempts to send a tx with utxos on account 1 + 2 and it fails. the whole point of the xp wallet balance is to mirror the send flow

gergelylovas:
> @frichards -- does signing work across the accounts for this implementation? it is not enough to just show a balance rolled up across the wallet for XP chain if the user cannot sign for all those UTXOS. it is a broken experience to return account 1 + 2 from getaddressesinrange and then a user attempts to send a tx with utxos on account 1 + 2 and it fails. the whole point of the xp wallet balance is to mirror the send flow
> 
> cc @gergelylovas

In extension the current UI does not allow to spend UTXOs across multiple accounts. This PR is only for fixing the displayed total balance. That being said we need to implement signing across multiple accounts where it's not supported like seedless and LL to provide full support for Web .

packages/service-worker/src/services/balances/handlers/getTotalBalanceForWallet/getTotalBalanceForWallet.ts
```diff
@@ -123,12 +123,38 @@ export class GetTotalBalanceForWalletHandler implements HandlerType {
...
 
       // Get balance for derived addresses
-      const { tokens: derivedAddressesBalances } =
-        await this.balanceAggregatorService.getBalancesForNetworks(
-          networksIncludedInTotal.map((network) => network.chainId),
-          derivedAccounts,
-          [TokenType.NATIVE, TokenType.ERC20],
-        );
+      // For P-chain, only fetch native AVAX tokens to ensure only native AVAX is included
+      const pChainNetworks = networksIncludedInTotal.filter(
```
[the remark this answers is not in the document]
frichards: The requirement doc says to "Only include P-Chain AVAX asset in balances". 
So I just tried to be very specific about it. I will revert this :D

vvava (approved):

PR #457, opened 2025-10-10
fix(balance): watchlist price

vvava:
## Description

We should fetch the prices for the tokens from a different endpoint at first try.

## Changes

Get the data from ```/watchlist/price``` before we use the prices from the ```/watchlist/tokens``` endpoint.

## Testing

Go to portfolio or network -> check the prices and compare them with web (they use ```/watchlist/price``` we should use similar values although the fetching time is different so small differences can be expected

## Checklist for the author


packages/service-worker/src/services/balances/BalanceAggregatorService.ts
```diff
@@ -202,24 +202,35 @@ export class BalanceAggregatorService implements OnLock, OnUnlock {
       !hasCurrentPrice ||
       (lastUpdated && lastUpdated + priceChangeRefreshRate < Date.now())
     ) {
-      const [priceChangesResult] = await resolve(
-        fetch(
-          `${process.env.PROXY_URL}/watchlist/tokens?currency=${selectedCurrency}`,
+      const [priceChangesResult, priceResult] = await Promise.allSettled([
```
csabbee: based on the `resolve` documentation in the sdk:
```typescript
const [result, error] = await resolve(promise)
if (result) return result
return handleError(error)
```

I think you don't need the `Promise.allSettled` (as a matter of fact I think it will always return `fulfilled` 😅 ) because that helper already handles the errors.

Something like this would do the trick I think:
```typescript
const [[priceChangesResult, priceChangeResultError], [priceResult, priceResultError]] = await Promise.all([
  resolve(
    fetch(
      `${process.env.PROXY_URL}/watchlist/tokens?currency=${selectedCurrency}`,
    ),
  ),
  resolve(fetch(`${process.env.PROXY_URL}/watchlist/price`)),
]);

if (priceChangeResultError) {
  return;
}

const priceChanges: PriceChangesData[] = await priceChangesResult.json();
const price = !!priceResultError ? {} : await priceResult.json()
```

csabbee (approved):

gergelylovas (approved):

PR #709, opened 2026-01-19
fix: enable P/X chain support for Keystone QR code connections

MoonBoi9001:
## Summary

Closes #708

The QR code scanner was only extracting the first key from Keystone devices, ignoring the Avalanche X/P key (`44'/9000'/0'`) even when present in Keystone 3 Pro QR data. This caused P/X chain operations to fail with "network not supported" errors.

- Extract all keys from QR codes, not just the first one
- Parse both EVM (`44'/60'`) and Avalanche (`44'/9000'`) derivation paths
- Set QR code as the default connection method
- Update UI text to clarify device support and add i18n
- Reorder dropdown to show QR option first

Gracefully falls back to EVM-only support if the QR code doesn't include an Avalanche key.

🤖 Generated with [Claude Code](https://claude.ai/code)


MoonBoi9001:
@gergelylovas @meeh0w wonder if you can take a look at this pr?

meeh0w:
@MoonBoi9001 I'll review & test next week 🙏

meeh0w (changes requested):
I only left two comments so far.

I did not manage to test it since my Keystone 3 Pro is now on a beta firmware which...doesn't work very well when trying to onboard using QR codes :)

I've contacted the Keystone team and will recheck this PR once they provide me with an updated firmware version.

meeh0w (commented):
Onboarding looks good, but it still doesn't allow for sending these funds using QR codes only.

So far we relied on the presence of Avalanche public keys to determine what kind of device was onboarded:

https://github.com/ava-labs/core-extension/blob/27541465706f730205893b4d02c2ef6596271958/packages/service-worker/src/services/onboarding/handlers/keystoneOnboardingHandlerNew.ts#L70-L76

Later on we relied on this info (`SecretType`) to determine if the user should use USB or QR codes to sign transactions.

With your changes we can no longer do that, since QR code will still extract the Avalanche keys, and so the UI will still prompt you to connect the device.

This is a good start (huge thank you! ❤️), but requires more changes.

Personally I would pause for a little bit, since Keystone team is about to release a new version of their firmware as they updated `@keystonehq/hw-app-avalanche`. With their changes, users will be able to Account 2 (and higher), as those were not usable for now (see https://github.com/ava-labs/core-extension/pull/739); and also sign transactions involving multiple addresses from single account.

We also need to rethink how to determine which connection type to use -- should we ask the user each time when they sign a transaction? Or maybe add a setting somewhere in the UI?

gergelylovas:
Appreciate the contribution. However as of Sept 1st 2026 we are sunsetting Keystone support in our wallets. Closing this PR.

MoonBoi9001:
>Appreciate the contribution. However as of Sept 1st 2026 we are sunsetting Keystone support in our wallets. Closing this PR.

This is really sad. Keystone was the most secure way to use core.

MoonBoi9001:
Separately, how will you support users that staked their avax via keystone in core and still have months to go before staking finishes? what path will there be to recover that avax back to the c chain?

PR #750, opened 2026-02-04
feat: improve Keystone USB error handling for homepage requirement

Qkin-Keystone:
## Description

Improves Keystone USB error handling when device is not on homepage. Adds clear user guidance and stops ineffective auto-retry behavior.

## Changes

- Added specific error detection for "not on homepage" state
- Enhanced UI to display actionable instructions when error occurs
- Disabled auto-retry when device needs manual navigation to homepage
- Propagated error state through approval flow for better UX

## Testing

1. Connect Keystone 3 Pro via USB, navigate away from homepage
2. Attempt transaction signing
3. Verify clear error message and instructions appear
4. Return device to homepage and verify transaction succeeds

## Screenshots:

N/A - Error handling improvements, no visual design changes

## Checklist for the author


meeh0w (changes requested):
This PR kind of goes against the grain if you look at other wallets' implementations:

1. With other devices, the `Approve on <<device>>` message is only shown in the UI when the device is in the correct state. This here allows the user to sign, only to then show an error message. Don't get me wrong -- it's better than just erroring out without any context, but the end result is very similar and still annoying -- the user has to manually click reject and retry.
2. The PR also relies on `ApprovalController.requestApproval()` throwing an error, which it should never do. We rely on this method to always resolve (albeit with `{ error }` property) in other places.
3. It also creates a very Keystone-specific if-clauses in places where I would not expect them. If we want to add this, we should create a util that is more generic (i.e. `isErrorRetryable`, or something along those lines, and allow users to retry signing without forcing them out of the approval flow).
4. This PR creates a constant `KEYSTONE_NOT_IN_HOMEPAGE_ERROR` in _our_ repository that is meant to represent the error message returned by an _external_ SDK/device. I'd say this is not the way to go -- your SDK should either export a specific error class or error code that we can test the error against. Otherwise this is very fragile (if you change the error message on the device, everything fails here).


I'd say let's leave it as it is for now, or add a way to detect the device being on the wrong screen _before_ sending the signing request to the device.

gergelylovas:
Closing due to inactivity

PR #871, opened 2026-03-30
fix: refresh atomic balance when balances update

frichards:
## Description

[CP-13435](https://ava-labs.atlassian.net/browse/CP-13435)

After a successful cross-chain transfer, Core Web cleared the stuck-funds banner while the extension could still show stale atomic totals until the popup was reopened. Balance polling updates the service worker’s atomic cache and emits `BalanceServiceEvents.UPDATED`, but `accountAtomicBalances` in `BalancesProvider` was only refreshed when the polling effect’s dependencies changed, not on every balance update event.

## Changes

- When balance `UPDATED` events arrive, dispatch `UPDATE_BALANCES` as before, then refetch atomic funds for the active account via `fetchAtomicBalanceForAccount`.
- Moved the subscription `useEffect` to sit after `fetchAtomicBalanceForAccount` so dependencies stay valid.

## Testing

1. Connect the extension to Core Web.
2. Run a C-Chain → P-Chain CCT from web (approve export and import in the extension).
3. After the transfer completes and web clears the stuck-funds UI, open the extension: the stuck-funds banner should not appear when atomic memory is clear.
4. Optionally confirm that a genuine stuck-funds case still shows the banner.

## Screenshots:

https://github.com/user-attachments/assets/f1fe1c5f-fb57-4aec-a374-6c3477df6a60

## Checklist for the author

[CP-13435]: https://ava-labs.atlassian.net/browse/CP-13435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui/src/contexts/BalancesProvider/BalancesProvider.tsx
+++ b/packages/ui/src/contexts/BalancesProvider/BalancesProvider.tsx
@@ -241,24 +241,6 @@ export function BalancesProvider({ children }: PropsWithChildren) {
   }, []);
 
   useEffect(() => {
-    const subscription = events()
-      .pipe(
-        filter(isBalancesUpdatedEvent),
-        map((evt) => evt.value),
-      )
-      .subscribe((balancesData) => {
-        dispatch({
-          type: BalanceActionType.UPDATE_BALANCES,
-          payload: balancesData,
-        });
-      });
-
-    return () => {
-      subscription.unsubscribe();
-    };
-  }, [events]);
-
-  useEffect(() => {
     dispatch({
       type: BalanceActionType.SET_LOADING,
       payload: true,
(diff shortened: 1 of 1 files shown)


vvava (approved):

meeh0w (approved):
This works, but I think we should revisit the `BalancesProvider` logic.

It's just a bit weird -- we `fetchAtomicBalanceForAccount(...)` after receiving a `BalanceServiceEvents.UPDATED` event...but this event already contains the atomic balances breakdown, we just never populate it when we dispatch the `BalanceActionType.UPDATE_BALANCES` action.

I haven't verified thoroughly, but it seems to me the `fetchAtomicBalanceForAccount()` function calls a service worker handler that aggregates the data we already have on the frontend and returns it back to us.

If that's the case, I guess we could do it all on the frontend, directly in the `BalanceServiceEvents.UPDATED` handler.

PR #902, opened 2026-04-22
chore: cleanup legacy bridge and swap  CP-14063

gergelylovas:
## Description

Cleans up legacy bridge and swap implementations.
Legacy unified bridge integration is limited to activity history augmentation
Fixes swap and bridge in core concierge
Fixes buy button in trending tokens
Migrates storing bridge dev env config to settings context

https://ava-labs.atlassian.net/browse/CP-14063

## Changes

Remove a lot of files

## Testing

- Full regression testing is necessary

## Checklist for the author


apps/next/src/pages/Concierge/hooks/useFunctions.ts
```diff
@@ -77,6 +82,119 @@ export const useFunctions = ({ setIsTyping, setInput }) => {
...
+      const gasSettings = networkFee
+        ? {
+            maxFeePerGas: networkFee.high.maxFeePerGas,
+            maxPriorityFeePerGas: networkFee.high.maxPriorityFeePerGas,
+          }
+        : undefined;
+
+      await manager.transferAsset({
+        quote,
+        gasSettings: { estimateGasMarginBps: 0, ...gasSettings },
```
meeh0w: Do we want no margin at all here? 👀

apps/next/src/pages/Concierge/model.ts
```diff
@@ -132,29 +132,28 @@ export const functionDeclarations: FunctionDeclaration[] = [
   },
   {
     name: 'bridge',
-    description: `Send a token from one network to another.`,
+    description: `Bridge a token from one network to another using the Fusion bridge. The user can optionally specify the source network; if omitted, the active network is used.`,
```
meeh0w, line 135: What is the active network though? I mean, there is no way to change it.

Out of scope for this PR, but maybe we should revisit and change the prompts here to ask the user about the source and target chains too for swap (e.g. `Swap USDC from C-Chain to Solana` or `Swap USDC on C-Chain to ETH on Base`) and completely get rid of the bridge command.

packages/common/src/feature-flags.ts
meeh0w, line 1: Want to also remove `SWAP` / `SWAP_C_CHAIN` / `SWAP_ETHEREUM` / `SWAP_SOLANA` / and `SWAP_FEES_JUPITER`?
As well as `BRIDGE` / `BRIDGE_ETH` and `BRIDGE_BTC`?

packages/service-worker/src/services/settings/handlers/setBridgeDevEnv.ts
meeh0w, line 1: Thanks for moving that 🙇

packages/common/src/feature-flags.ts
gergelylovas: definitely! pushed [e404733](https://github.com/ava-labs/core-extension/pull/902/commits/e404733b1e2c206779273f1ce56af2f854b8e4da)

apps/next/src/pages/Concierge/model.ts
gergelylovas: Yes definitely. We still have the `network` property in the network context which gets injected into the system prompt replacing `__CURRENT_NETWORK_ID__`. Which is Avalanche C-chain by default.

apps/next/src/pages/Concierge/hooks/useFunctions.ts
gergelylovas: fixed, now using the same feature flag value as fusion elsewhere

meeh0w (approved):

bskyn (approved):

PR #905, opened 2026-04-23
fix: do not encrypt chain id

bohdan-struk-avalabs:
## Checklist for the author

### DIFF
--- a/apps/next/src/pages/Settings/components/Home/Home.tsx
+++ b/apps/next/src/pages/Settings/components/Home/Home.tsx
@@ -67,7 +67,7 @@ export const SettingsHomePage = () => {
   const { contacts } = useContactsContext();
   const { path } = useRouteMatch();
   const { push } = useHistory();
-  const { capture } = useAnalyticsContext();
+  const { capture, initAnalyticsIds, isInitialized } = useAnalyticsContext();
   const { featureFlags, isFlagEnabled } = useFeatureFlagContext();
 
   const {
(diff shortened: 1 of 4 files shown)


apps/next/src/pages/Settings/components/Home/Home.tsx
```diff
@@ -337,6 +337,9 @@ export const SettingsHomePage = () => {
                 capture('AnalyticsConsentSettingChanged', {
                   analyticsConsent: newValue,
                 });
+                if (newValue && !isInitialized) {
```
bskyn, line 338: So I traced this call and think there might be an issue.

`isInitialized` seems to be always stuck at false so this guard will always fall into and run `initAnalytics(true)` when you toggle which would re-initalizes a new user id

Take a look at `AnalyticsProvider` and the useEffect

```tsx
  .subscribe((ids) => {
    if (!ids) {
      setIsInitialized(!!ids);   // only enters when ids is falsy
    }
  });
```

The if (!ids) wrapper means the setter can only ever be called with a falsy value and `setIsInitialized` is unreachable. Maybe this was an existing bug but it didnt seem to affect anything because the `isInitialized` state from the analytics provider wasn't used anywhere until now?

meeh0w: Nice catch, fixed.

meeh0w (approved):

LMulvey (approved):

bskyn (approved):

PR #908, opened 2026-04-24
chore(e2e): optimize CI — alpha release for regression, dev build for smoke

blockqa:
## Description

Optimizes E2E CI by avoiding unnecessary builds and aligning each suite with the artifact it should validate:

- **Regression tests** now download the latest alpha release zip from GitHub Releases instead of rebuilding from source. Saves CI time and exercises the actual shipped artifact.
- **Smoke tests** now use a full dev-mode build across the entire monorepo (frontend + packages), matching local `yarn dev` behavior so the extension environment the team runs locally is what gets smoke-tested.

## Changes

- `.github/workflows/e2e_regression_tests.yaml`: replaced the build job with a download step that fetches the latest alpha release via `gh release download`.
- `.github/workflows/e2e_smoke_tests.yaml`: switched to dev build using `build:dev:no-source-maps`, with `DEV_*` prefixed repository secrets injected into `.env.dev` (including `DEV_ID_SERVICE_API_KEY` and `DEV_GLACIER_API_KEY`, which the dev backends require).
- Root `package.json`: added `build:packages:dev` and `build:dev:no-source-maps`; `build:dev` now fans out dev configs across all packages.
- Package `package.json`s (service-worker, content-script, offscreen, inpage): added `build:dev` scripts so each package can be built in dev mode standalone.
- Dev `rsbuild.*.ts` configs: honor `NO_SOURCE_MAPS=true` to keep CI bundles small (inline source maps were crashing the extension in CI).
- `e2e/helpers/loadWalletSnapshot.ts`: reads snapshots from `storage-snapshots/${SNAPSHOT_SET}/` (defaults to `alpha` locally), with a fallback to the flat root for backward compatibility.
- `e2e/.gitignore`: recursively ignores snapshot files under `storage-snapshots/**`.
- Workflows set `SNAPSHOT_SET=alpha` (regression) and `SNAPSHOT_SET=dev` (smoke) and sync from `s3://.../ext/alpha/` or `s3://.../ext/dev/` into the matching local subdirectory.

## Testing

- Regression CI will download the latest alpha zip and run against it.
- Smoke CI will produce a dev build using `DEV_*` secrets and load snapshots generated with the Dev public key.
- Verified dev build succeeds locally with `yarn build:dev:no-source-maps`.

## Checklist for the author


ryanewood (approved):
lgtm, curious if we should build the extension any differently in web CI? right now we use `yarn build:alpha`, but haven't run into any problems yet.

meeh0w (approved):

PR #913, opened 2026-04-29
[CP-14144] exclude spam tokens properly

bferenc:
## Description
The "hide tokens flagged as spam" toggle on the "manage tokens" page only excluded `malicious` tokens. This change makes it to properly exclude anything that's not `benign`, `warning` or unknown just like we do during the balance calculation.

## Changes
Properly exclude anything that's not `benign`, `warning` or unknown on the token list

## Testing
Verify that visibility for `spam` tokens can be toggled as expected.

## Checklist for the author


meeh0w (approved):

blockqa (approved):
Looks much better now. I checked against the explorers, and we are not filtering valid tokens. Etherscan seems a bit more restrictive with low-reputation tokens, but given the current setup, this looks good to me.

PR #924, opened 2026-05-09
feat(core): integrate isVerified into swap token list

LMulvey:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14123
Figma: <could-not-find>

This PR modifies Swap inside of Core Extension to take advantage of the `isVerified` field on the `/tokenlist` endpoint. It functions similarly to the Core Web implementation.

- On Swap, display token dropdown normally.
- If the user filters token dropdown via search, separate `isVerified: false` tokens to the bottom of the list and visually separate with `Unverified tokens` divider

## Notes

- I could not find an extension ticket or specific designs so these are _best guesses_. Will update on Monday if proper ones are found but these are likely correct based on the current design language in the Extension.
- Much like Web, the `isVerified` only exists on C-Chain right now. For other chains, it is returned as `null` which we are treating as `true` right now.

## Testing

Test must be done with local build of extension.

## Testing Instructions

1. Add credentials/login to extension
2. Visit swap page
3. Click dropdown for destination
4. See no `Unverified tokens` divider
5. Add search string
6. See `Unverified tokens` divider - scroll past and see it stick to the top of the list.
7. Spot-check `Unverified tokens` for accuracy!
8. Change chain to Ethereum
9. See no `Unverified tokens` divider
10. Add search string
11. See no `Unverified tokens` divider

## Media

Screenshot 2026-05-08 at 5 57 19 PM
Screenshot 2026-05-08 at 5 57 29 PM


bohdan-struk-avalabs (dismissed):
Code looks good.

apps/next/src/pages/Fusion/contexts/hooks/useSwapTargetTokenList.ts
```diff
@@ -40,22 +41,28 @@ export const useSwapTargetTokenList = (
...
-      );
+    } else {
+      filtered = allTokens;
     }
 
-    return allTokens;
+    const verified = filtered.filter((t) => t.isVerified !== false);
+    const unverified = filtered.filter((t) => t.isVerified === false);
+
+    return [...verified, ...unverified];
```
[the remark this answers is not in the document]
LMulvey: intended

bohdan-struk-avalabs (dismissed):

apps/next/src/components/TokenSelect/TokenSelect.tsx
```diff
@@ -100,4 +143,286 @@ function TokenSelectRaw({
...
+      (token) => token.isVerified !== false,
+    ).length;
+    const hasUnverified = displayTokens.some(
+      (token) => token.isVerified === false,
+    );
+    const sep = hasUnverified ? verifiedCount : -1;
+    return {
+      separatorAt: sep,
+      totalItems: displayTokens.length + (sep >= 0 ? 1 : 0),
+    };
```
erictaylor, line 212: Is `displayTokens` already correctly ordered (ie verified, then unverified)?

Is there a potential problem here where the index could be incorrect because of the order?

---
Example Problem:

Say `displayTokens` is `[U1, V1, V2, U2]` (U = Unverified, V = Verified).

Code computes:
* `verifiedCount` = 2
* `hasUnverified` = true
* `separatorAt` = 2

What gets rendered:
-  rows before separator (index 0,1): U1, V1
- separator row at index 2: “Unverified tokens”
- rows after separator: V2, U2

So you end up with:
- an unverified token above the unverified header (U1)
- a verified token below the unverified header (V2)
---

apps/next/src/components/TokenSelect/TokenSelect.tsx
```diff
@@ -100,4 +143,286 @@ function TokenSelectRaw({
...
+    setSelectedChainId,
+  };
+
+  if (tokenList.some((t) => t.isVerified === false)) {
+    return <TokenSelectWithSeparator {...sharedProps} />;
+  }
+  return <TokenSelectFlat {...sharedProps} />;
+}
+
 export const TokenSelect = memo(TokenSelectRaw, areTokenListsEqual);
```
erictaylor, line 432: `TokenSelect` is memoized with `areTokenListsEqual`.

See line 422 comment.

apps/next/src/components/TokenSelect/TokenSelect.tsx
```diff
@@ -100,4 +143,286 @@ function TokenSelectRaw({
...
+    ...props,
+    filteredTokenList,
+    chainOptions,
+    selectedChainId,
+    setSelectedChainId,
+  };
+
+  if (tokenList.some((t) => t.isVerified === false)) {
+    return <TokenSelectWithSeparator {...sharedProps} />;
+  }
```
erictaylor, line 428: memoization with `areTokenListsEqual` comparator only compares token ids and not `isVerified`.

This could lead to this line not re-evaling correctly.

---

Example:

1. Initial props:
    - tokenList = [AVAX(verified), USDC(verified)] (V = Verified Token)
    - branch picks TokenSelectFlat (no unverified tokens)
2. Refresh arrives:
    - tokenList = [AVAX(verified), USDC(unverified)] (same IDs, only isVerified changed)
3. Comparator returns true (IDs unchanged), so TokenSelectRaw does not re-render.
4. Branch is not re-evaluated, so UI stays flat (no separator), even though USDC is now unverified.

It only updates later if some unrelated prop changes.

erictaylor (dismissed):

apps/next/lavamoat/rspack/policy.json
```diff
@@ -307,7 +309,6 @@
         "@core/ui>i18next": true,
         "@core/service-worker>joi": true,
         "@core/ui>lodash": true,
-        "@rsbuild/plugin-node-polyfill>process": true,
```
erictaylor, line 310: Maybe we need to add this back in?

apps/next/lavamoat/rspack/policy.json
```diff
@@ -405,7 +406,6 @@
         "@core/ui>ledger-bitcoin": true,
         "@core/ui>lodash": true,
         "@core/ui>micro-signals": true,
-        "@rsbuild/plugin-node-polyfill>process": true,
```
erictaylor, line 408: Add back in?

erictaylor (approved):

bohdan-struk-avalabs (approved):

PR #934, opened 2026-05-18
fix: seedless P-chain signing on non-zero accounts (CP-14284)

B0Y3R-AVA:
# Summary

Jira ticket: [CP-14284](https://ava-labs.atlassian.net/browse/CP-14284)
Figma: n/a

Fixes a regression introduced by [#746](https://github.com/ava-labs/core-extension/pull/746) where seedless wallets on any non-zero account index can't sign P-chain / keychain transactions. The dapp surfaces *"can't sign message"* and the approval popup keeps reappearing.

That PR repurposed `accountIndices` in `WalletService.#getSeedlessWallet` from BIP-44 account indices to **address indices within the active account**, but the fallback was left as `[accountIndex ?? secrets.account.index]`. For the single-signer path (no explicit `externalIndices`) this puts the active account's BIP-44 index into the *address* segment, producing `m/44'/9000'/<acct>'/0/<acct>` — a path CubeSigner never provisions. It worked on account 0 only because both segments happen to be `0`.

The fix defaults `addressIndices` to `[0]` for the single-signer path so the derivation path becomes `m/44'/9000'/<acct>'/0/0`, which matches what CubeSigner holds.

## Notes

The bug never surfaced in Sentry — `SeedlessWallet.#handleError` is a bare rethrow and none of the seedless signing throw sites are wrapped in `Monitoring.sentryCaptureException`. To make sure the next regression in derivation-path resolution is observable, this PR also instruments:

- `WalletService.#getSeedlessWallet` — both `'No public keys available'` and `'Some of requested signing keys are not available'` throws.
- `SeedlessWallet.#handleError` — non-token-expired errors (covers `signAvalancheTx`, `signTx`, `signTransaction`, `signMessage`, `#getSigningKey`, `#signBlob`, `#signEip191`, `signSolanaTx`).
- `SeedlessWallet.signMessage(AVALANCHE_SIGN)` precheck throws (`'Public key not available'`, `'X/P public key not available'`) that bypass `#handleError`.

Token-expired errors keep flowing through `sessionManager.notifyTokenExpired()` unchanged.

Also corrects the stray `'No public keys not available'` message string (double negation typo) introduced by the original code.

Mobile (`core-mobile`) was verified unaffected — its `SeedlessWallet` keeps `accountIndex` and `addressIndex` as distinct params throughout, so the same conflation can't happen.

## Testing

- Seedless wallet (CubeSigner / Cube Signer) with at least two derived accounts (indices 0 and 1+).
- Mainnet or Fuji P-chain — easiest repro is the Builders Hub L1 validator balance flow:
  https://build.avax.network/console/layer-1/l1-validator-balance
- No special local storage overrides needed; behavior changes purely server-side of the extension.

## Testing Instructions

1. Set up a seedless wallet and add account index 1 (and optionally 2+).
2. Switch to account index 0 — initiate an L1 validator balance increase → confirm the tx signs and submits as before (regression guard).
3. Switch to account index 1 (or 2+) — initiate the same flow → confirm the tx signs and submits successfully (previously failed with *"can't sign message"* and a repeating approval popup).
4. While on account 1, sign an X-chain off-chain message (e.g. via a dapp that issues `avalanche_signMessage`) — confirm signature succeeds.
5. While on account 1, run a Fusion swap on C-chain to confirm EVM signing is still healthy.

Automated coverage:
- New unit test `WalletService.test.ts > sign > avalanche signing - XP / Coreth > on a non-zero active account with no external indices` asserts the derivation path is built with addressIndex=0 (not the active account index) and that the right pubkey ends up in the constructed `SeedlessWallet`.
- Multi-address case (`externalIndices: [0, 1]`) test from #746 still passes.

## Media

confirmed testing steps with Owen who reported this issue. 

https://github.com/user-attachments/assets/a2a8a219-2a01-440d-bcf3-dec9e709210c

[CP-14284]: https://ava-labs.atlassian.net/browse/CP-14284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/service-worker/src/services/seedless/SeedlessWallet.test.ts
+++ b/packages/service-worker/src/services/seedless/SeedlessWallet.test.ts
@@ -60,6 +60,10 @@ jest.mock('./SeedlessMfaService');
 jest.mock('@core/common', () => ({
   ...jest.requireActual('@core/common'),
   getProviderForNetwork: jest.fn(),
+  Monitoring: {
+    ...jest.requireActual('@core/common').Monitoring,
+    sentryCaptureException: jest.fn(),
+  },
 }));
 jest.mock('@scure/base');
 
--- a/packages/service-worker/src/services/seedless/SeedlessWallet.ts
+++ b/packages/service-worker/src/services/seedless/SeedlessWallet.ts
@@ -34,6 +34,7 @@ import {
   hasAtLeastOneElement,
   isBitcoinNetwork,
   isTokenExpiredError,
+  Monitoring,
 } from '@core/common';
 import {
   AddressPublicKeyJson,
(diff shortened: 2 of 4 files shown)


erictaylor (approved):
Looks good to me.

bskyn (approved):

PR #939, opened 2026-05-27
ci: lavamoat policy check+update workflow

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14348
Figma: _not applicable_

1. Sets up CI infrastructure to keep LavaMoat policy files in sync without requiring contributors to run production builds locally. Adds a workflow that detects policy drift on every PR, and a comment-triggered companion that regenerates policies in CI and pushes the fix back to the PR branch as a signed (Verified) commit.

2. Replaces the inline `yarn build` + drift check that used to live in `pr-checks.yaml` with two dedicated workflow files:
- **`validate-lavamoat-policies.yaml`** — runs `yarn build` and diffs every `**/lavamoat/rspack/policy.json` against the committed copy. When the policy drifts it saves a `.patch` artifact and fails the check.
- **`update-lavamoat-policies.yaml`** — triggered by a `/update-lavamoat-policies` comment on a PR. Finds the failed validate run, downloads the diff artifact, applies it, creates a signed commit via the GitHub GraphQL `createCommitOnBranch` mutation (same server-side signing path `dependabot[bot]` and `changesets/action` use in `github-api` mode), and re-dispatches `validate-lavamoat-policies` against the new HEAD so the previously red check turns green on its own.

## Notes
**Why CI instead of local regeneration.** The LavaMoat plugin inlines `process.env.*` literals it sees in source into the generated policy. Without production secrets locally, regenerated policies drift on the env-var diff alone - contributors would either need to hand-edit policy files or hold prod secrets on their dev machines. CI has the secrets already, so this moves the regeneration burden there.

**Why a separate validate workflow.** The validate workflow is its own top-level workflow (rather than a `workflow_call` job inside `pr-checks.yaml`) so the update workflow can re-dispatch *only* it via `gh workflow run`. 

Re-dispatching the full `pr-checks` would re-run lint/typecheck/scanner/test on a `policy.json`-only commit, which is pure waste — a `policy.json` diff cannot break those checks.

⚠️ 
**Branch protection follow-up.** This PR removes the old `Build extension` + inline drift check from `pr-checks`'s `Lint` job. After merge, the branch protection required-checks list needs to be updated to:
- `Lint` (from `pr-checks.yaml`)
- `Validate LavaMoat policies` (from `validate-lavamoat-policies.yaml`)

## Testing

`issue_comment` workflows only fire for workflows already on the default branch, so `update-lavamoat-policies` cannot be exercised on this PR itself. The validate side IS testable here (it uses `pull_request`, which loads from the PR branch).

### On this PR before merge
1. Confirm the `Validate LavaMoat policies` check ran on this PR,   executed `yarn build`, detected no drift, and exited successfully.

### After merge
1. Update branch protection on `main` to require `Lint` and  `Validate LavaMoat policies` (and drop any stale `Lint and build` requirement).
3. Open a throwaway PR that intentionally edits a committed `apps/next/lavamoat/rspack/policy.json` (e.g. delete one entry from a resource's `globals` map).
4. Confirm `Validate LavaMoat policies` fails on that PR and the failed run's Artifacts section contains `lavamoat-policy-diff`.
5. Comment `/update-lavamoat-policies` on the PR.
6. Within a minute, expect:
   - 👍 reaction on the comment.
   - A new commit on the PR branch authored by `github-actions[bot]`, subject `chore: update LavaMoat policies`, showing the green **Verified** badge.
   - A new `validate-lavamoat-policies` run on the new HEAD that turns the check green.
   - A success comment from the bot.
7. Open the new commit and confirm it only touches `policy.json` files
   and the contents match what local `yarn build` would have produced.

## Media
_not applicable_

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -2,7 +2,7 @@ name: PR checks
 on: pull_request
 jobs:
   lint:
-    name: Lint
+    name: Lint and build
     runs-on: ubuntu-latest
     environment: alpha
     steps:
@@ -39,12 +39,3 @@ jobs:
           if [[ $(git status --short) == '' ]]; then exit 0; else exit 1; fi
       - name: Test
         run: yarn test
-
-  # Production build + LavaMoat policy drift detection live in their own
-  # reusable workflow. It uploads a `.patch` artifact when policies drift,
-  # which `update-lavamoat-policies.yaml` replays back onto the PR branch
-  # when a maintainer comments `/update-lavamoat-policies`.
-  validate-lavamoat-policies:
-    name: Validate LavaMoat policies
-    uses: ./.github/workflows/validate-lavamoat-policies.yaml
-    secrets: inherit
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -2,7 +2,7 @@ name: PR checks
 on: pull_request
 jobs:
   lint:
-    name: Lint and build
+    name: Lint
     runs-on: ubuntu-latest
     environment: alpha
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -39,18 +39,3 @@ jobs:
           if [[ $(git status --short) == '' ]]; then exit 0; else exit 1; fi
       - name: Test
         run: yarn test
-      - name: Build extension
```
meeh0w, line 42: Build step gets moved to `validate-lavamoat-policies`. The only reason it's here is to check if the app builds; it can very well be done in the other workflow where the build output will actually be used (regenerated policies will be used to determine if they need an update).

LMulvey (dismissed):
LGTM

csabbee (approved):
Wow, I know something must be wrong with me, but I like a nice and thought through CI setup 😅

LMulvey (approved):

PR #942, opened 2026-06-02
fix: defer initialization to avoid brave renderer crash

LMulvey:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14355
Figma: N/A

This PR attempts to hotfix a crash introduced in Brave v1.90.128 (See: https://github.com/brave/brave-core/pull/36564/changes).

The TL;dr non-computer science-y reason for the crash was that a change was made to Brave's Wallet code to no longer use a raw pointer for the `RenderFrame` reference used in the code. It instead uses an existing reference than appears to sometimes be stale. When the extension announces its presence via `EIP6963`, it triggers a memory fault as the reference becomes stale. This appears to be fixed when differing the announcement by a few frames and giving the code more of a chance to stabilize. 

In testing, this introduction of a timeout appears to have no meaningful effect to the end user (besides the _positive_ effect of allowing extension to properly load on Brave).

## Testing

1. Download/launch Brave >=1.90.128
2. Install or have Core Extension
3. Ensure Brave Wallet is enabled and added as an Ethereum fallback on your Web3 settings (`brave://settings/web3`)
4. Ensure that you can navigate web pages as expected (previous crash would happen almost instantaneously)
5. Ensure you can connect to Core (or any dApp) and validate functionality
6. Ensure crash does not occur

### DIFF
--- a/packages/inpage/src/initializeInpageProvider.ts
+++ b/packages/inpage/src/initializeInpageProvider.ts
@@ -51,25 +51,35 @@ export function initializeProvider(
   );
   const multiWalletProxy = createMultiWalletProxy(evmProvider);
 
-  globalObject.addEventListener('eip6963:announceProvider', (event: any) => {
-    multiWalletProxy.addProvider(
-      new Proxy(
-        {
-          info: { ...event.detail.info },
-          ...event.detail.provider,
-        },
-        {
-          deleteProperty: () => true,
-          set: () => true,
-        },
-      ),
-    );
-  });
-
   setGlobalProvider(evmProvider, globalObject, multiWalletProxy);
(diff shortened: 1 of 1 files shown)


erictaylor (changes requested):
Couple high level thoughts:

1. Do we need to delay setting up the `announceProvider` provider listener? Our own listener shouldn't effect their problem right? Don't we just need to delay our firing of events (not necessarily our listeners)?
2. I wonder if it would make more sense and work to not use `setTimeout` with an arbitrary 500ms delay, but instead to use something like `requestIdleCallback`, or `queueMicrotask`?

LMulvey:
Closing in favor of #943

PR #943, opened 2026-06-02
fix: brave wallet conflicts

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14355
Figma: _n/a_

## Notes
After some debugging I think it's the `...event.details.provider` spread that's causing issues.

My suspicion is that the Brave Wallet is also a `Proxy` object, it has some recursive `get` trap set and when we try to spread it, it hangs the tab.

## Testing
1. Open Brave with Brave Wallet and Core Extension enabled and navigate to any pages/dApps.
2. The browser should not crash.

## Media
_n/a_


LMulvey (dismissed):

erictaylor (dismissed):

meeh0w:
@copilot review again

LMulvey (approved):

ryanewood (approved):

PR #946, opened 2026-06-03
ci: improve build workflows security

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-13618
Figma: _n/a_

⚠️ **This intentionally _does not_ apply the `@lavamoat/node` protections to our build scripts.**

## Out of scope: build-time sandboxing via `@lavamoat/node`
This PR layers in install-time integrity (`yarn install --check-cache`)
and narrows the secret scope of `pr-checks`, but it intentionally does
**not** wrap the build process itself in `@lavamoat/node`. Filing here
so the decision is reviewable; the corresponding ticket gets a comment with the same rationale.

### What `@lavamoat/node` would buy us
It sandboxes every build-time dependency (`@rspack/core`, `@swc/core`,
every babel plugin, etc.) in its own SES Compartment with policy-
restricted access to `globalThis`, `fs`, `net`, `child_process`, and
the rest of Node's surface. In principle this prevents a compromised
build dep from exfiltrating `.env.production` secrets, rewriting the
bundle mid-codegen, or reaching out to attacker infrastructure during
`yarn build`.
### Why we're not doing it
**1. The protection has a structural gap we can't close without changing the bundler itself.** Rspack runs
codegen in Rust via N-API. SES Compartments sandbox JavaScript
execution; they don't reach into native bindings. A compromise of
`@rspack/binding-*`'s Rust binary would let an attacker rewrite bundle
output before LavaMoat ever sees it, regardless of any build-time
policy we ship. The largest realistic attack vector at the build layer
is exactly the one `@lavamoat/node` can't address.

**2. Policy maintenance has a known failure mode at our team size.**
Build-time policies are dominated by mechanical entries — "rspack needs
`fs.realpath`", "babel needs `Module.createRequire`". 95%+ of every
diff is "yes, of course this build tool needs that, approve." Without
a reviewer who has a mental model of what each transitive dep should
legitimately be allowed to do — i.e. a supply-chain-specialised role
we don't have on our team of 5 — the review likely degrades to rubber-stamping
very quick. At that point the gate produces **negative**
security: a green check that signals nothing, plus reviewer attention
that could go somewhere else.

For scale reference: our runtime LavaMoat policies — which are smaller
and more focused than a build-time policy would be, with a threat model
("don't leak the seed phrase") the team can actually reason about —
already weigh 1.1k–6.6k lines per bundle. A build-time policy covering
the entire dev-dep tree would be substantially larger and substantially
less tractable to review.

**3. The marginal protection over what we already have is small.** The
existing stack covers most of the realistic build-time attack surface:
- `enableScripts: false` + `@lavamoat/allow-scripts` → no lifecycle
  scripts execute during install
- `yarn install --check-cache` *(this PR)* → tarball checksums
  re-validated against the registry on every install
- Signed-commits-only branch protection + reviewed `yarn.lock` → no
  silent lock mutation
- `semgrep-ava-labs[bot]` with reachability analysis → CVE detection
  on every PR
- `pr-checks` secret-scope narrowing *(this PR)* → no prod secrets
  reach lint / typecheck / test
- Runtime LavaMoat on every shipping bundle → if a malicious dep does
  make it past install-time gates, its in-browser capabilities are
  policy-restricted
The slice `@lavamoat/node` would close (a build-time dep exfiltrating
secrets or subtly modifying bundle output in ways the runtime policy
doesn't catch) is real but narrow.

## Testing
1. CI passing

## Media
_n/a_

### DIFF
--- a/.github/workflows/create_release.yaml
+++ b/.github/workflows/create_release.yaml
@@ -52,9 +52,13 @@ jobs:
           echo 'TOKEN_AGGREGATOR_SERVICE_URL="${{ secrets.TOKEN_AGGREGATOR_SERVICE_URL }}"' >> .env.production
           echo 'MARKR_API_URL="${{ secrets.MARKR_API_URL }}"' >> .env.production
           echo 'MARKR_API_TOKEN="${{ secrets.MARKR_API_TOKEN }}"' >> .env.production
+      # See `main_branch.yaml` for the rationale on `--check-cache`.
+      # Production builds ship to end users with non-custodial wallet
+      # keys, so the same belt-and-suspenders applies — and arguably
+      # more so here.
       - name: Install dependencies
         run: |
-          yarn install
+          yarn install --check-cache
           yarn allow-scripts
       - name: Build library
         run: yarn build
--- a/.github/workflows/main_branch.yaml
+++ b/.github/workflows/main_branch.yaml
@@ -55,9 +55,13 @@ jobs:
           echo 'TOKEN_AGGREGATOR_SERVICE_URL="${{ secrets.TOKEN_AGGREGATOR_SERVICE_URL }}"' >> .env.production
           echo 'MARKR_API_URL="${{ secrets.MARKR_API_URL }}"' >> .env.production
           echo 'MARKR_API_TOKEN="${{ secrets.MARKR_API_TOKEN }}"' >> .env.production
+      # `--check-cache` re-validates every tarball checksum against the
+      # registry on each install instead of trusting a possibly poisoned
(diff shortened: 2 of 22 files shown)


erictaylor (dismissed):
LGTM. An additional workflow hardening we could do in most workflows would be to add the `persist-credentials: false` option to `actions/checkout`.

LMulvey (dismissed):

csabbee (approved):

LMulvey (approved):

PR #963, opened 2026-06-26
feat: recurring swap notifications

meeh0w:
## Summary

⚠️ This PR is based on #962

**Ticket:** https://ava-labs.atlassian.net/browse/CP-14590

Adds push notifications for recurring (DCA) swaps. The device subscribes to per-order updates and surfaces them both as native OS notifications and as entries in the in-app notification center. All discovery and subscription is owned by the service worker, so it works even if the popup is closed.

## What's included

**Service worker (background-owned subscription)**
- New `RecurringSwapNotificationService`:
  - Listens for recurring swaps notifications
  - **Discovers** the active account's orders via the existing Fusion `TransferManager` (`TransferTrackingService.getManager()`), and subscribes to active/paused order ids
  - **`chrome.alarms`-driven polling** rides out Markr's indexing lag after creation, then self-stops when there's no outstanding work — no indefinite polling. Survives popup close and SW restarts (listener re-registered in `BackgroundRuntime.activate()`).
- `NOTIFICATION_DISCOVER_RECURRING_SWAPS` handler — the UI fires this once after creating an order to hand discovery off to the SW.

**Gating on the Balance changes toggle**
- Recurring swaps reuse the single "Balance changes" preference.
- When the toggle is turned **off**: local subscription state is cleared and discovery is stopped. Since the backend has no unsubscribe endpoint, incoming pushes are additionally **filtered out** while disabled. Re-enabling re-subscribes.

**Notification center UI**
- `RECURRING_SWAP` notification type, metadata + status schema (`active`/`completed`), type guard, and transform support.
- `RecurringSwapItem` row with status label, and a `NotificationIcon` update (recurring icon + chain-logo badge for the chain the swap occurred on).

## How it works

1. User creates a recurring swap → UI sends `NOTIFICATION_DISCOVER_RECURRING_SWAPS` to the SW.
2. SW opens a watch window and polls Markr until the new order is indexed, then subscribes it.
3. Order executes/completes → backend pushes via FCM → SW shows a native toast and adds a notification-center entry.

## Testing instructions

1. Enable the recurring swaps feature flag and "Balance changes" notifications.
2. Create a short recurring swap (C-Chain mainnet) and close the popup.
3. Confirm a native notification arrives on execution/completion and an entry appears in the notification center with the correct status + chain badge.
4. Toggle "Balance changes" off → confirm no further recurring-swap notifications surface; toggle back on → confirm re-subscription.

## Media

image

image


erictaylor (dismissed):
Code review looks good. Did not test.

csabbee (approved):

0xstt (approved):

PR #965, opened 2026-06-30
feat: add Avalanche CCT support to Fusion swaps

erictaylor:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14206
Figma: N/A

Adds Avalanche CCT support to Fusion swaps behind the `fusion-avalanche-cct` feature flag.

- Wires the Fusion SDK `AVALANCHE_CCT` service initializer with Avalanche wallet functions for sending C/P/X transactions, loading UTXOs, resolving wallet addresses, and deriving change addresses.
- Updates Fusion source and destination token selectors to expose C-Chain, P-Chain, and X-Chain AVAX options with the expected network pill ordering and 9-decimal native AVAX on P/X chains.
- Formats CCT quote addresses with the expected C/P/X prefixes and supports 0-value recovery quotes for stuck atomic UTXO imports.
- Hides slippage and price impact for services where they are not applicable.
- Adds the two-transaction CCT warning only for positive-value Avalanche CCT quotes.
- Suppresses generic optimistic success toast/confetti for Avalanche CCT export/import transaction prompts so success is shown after transfer completion instead.
- Ensures the background transfer tracker initializes with the Avalanche CCT service so CCT transfers can progress from destination pending to completed.

## Notes

Avalanche CCT quote and transfer support is intentionally feature-gated. The UI treats C/P/X AVAX as separate selectable network assets, but only native AVAX is exposed for P-Chain and X-Chain.

Recovery quotes are allowed to request a `0` input amount so the Fusion SDK can surface the appropriate minimum amount behavior or complete a pending import-only recovery flow.

Specific UTXO selection is _not_ included in this PR work.

## Testing Instructions

1. Enable the `fusion-avalanche-cct` feature flag.
2. Open Fusion swap with an account that has native AVAX on Avalanche C/P/X chains.
3. Verify the source and destination token selectors show separate C-Chain, P-Chain, and X-Chain filter pills, with C-Chain immediately followed by P-Chain and X-Chain.
4. Select a positive-value CCT route, such as C-Chain AVAX to P-Chain AVAX.
5. Confirm the warning explains that two transactions are required.
6. Submit the swap and verify the wallet prompts for export first, then import.
7. Verify generic transaction success toast/confetti is not shown after the export prompt and the transfer progress screen is shown after the import prompt.
8. Verify the transfer tracker progresses the destination side from pending to completed.
9. Test a `0` input amount CCT recovery quote path and verify the two-transaction warning is not shown.

## Media

https://github.com/user-attachments/assets/4d6fc84e-3603-4250-86bf-412a6b27b4ef


apps/next/src/pages/Fusion/contexts/FusionStateContext.tsx
```diff
@@ -142,11 +149,7 @@ export const FusionStateContextProvider: FC<{ children: ReactNode }> = ({
...
 
-  const isAmountLowerThanMinimum =
-    typeof minimumTransferAmount === 'bigint' &&
-    sourceAmountBigInt < minimumTransferAmount;
-
-  const skipFetching = isAmountHigherThanBalance || isAmountLowerThanMinimum;
+  const skipFetching = isAmountHigherThanBalance;
 
   // Avoid spamming quoters by debouncing the user amount
   const [debouncedUserAmount] = useDebouncedValue(userAmount, {
```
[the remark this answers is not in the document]
erictaylor: Addressed in the latest push. Empty amount input now maps to `undefined`, so quote streaming is skipped when the field is cleared, while an explicit `0` still maps to `0n` for CCT recovery quotes.

apps/next/src/components/TokenSelect/TokenSelect.tsx
```diff
@@ -399,21 +399,26 @@ function TokenSelectWithSeparator({
...
     tokenList,
     selectedChainId,
     isAnyAvalancheNetwork,
   );
-  const chainOptions = useChainOptions(availableChainIds, hasAvalancheNetworks);
+  const chainOptions = useChainOptions(
+    availableChainIds,
+    hasAvalancheNetworks,
+    chainFilterMode,
+  );
```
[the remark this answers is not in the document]
erictaylor: Addressed in the latest push. The selector now treats stale chain selections as `null`, resets the state when the current selection is no longer in `chainOptions`, and includes `chainFilterMode` in the memo comparator so feature-flag mode changes re-render correctly.

apps/next/src/pages/Fusion/contexts/FusionStateContext.tsx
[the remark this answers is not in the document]
erictaylor: @copilot did e100501 address the issue? If so please mark this review comment as resolved. If not, please leave a comment detailing the continued issue.

apps/next/src/components/TokenSelect/TokenSelect.tsx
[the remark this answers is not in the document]
erictaylor: @copilot did e100501 address the issue? If so please mark this review comment as resolved. If not, please leave a comment detailing the continued issue.

apps/next/src/components/TokenSelect/hooks.ts
```diff
@@ -108,11 +115,62 @@ export const useChainOptions = (
...
     });
 
     return options;
-  }, [availableChainIds, hasAvalancheNetworks, getNetwork]);
+  }, [availableChainIds, hasAvalancheNetworks, getNetwork, chainFilterMode]);
+};
+
+const sortChainIds = (a: number, b: number) => {
+  const aPriority = getAvalancheChainPriority(a);
+  const bPriority = getAvalancheChainPriority(b);
```
bohdan-struk-avalabs: Can you please add those functions into separate utils file, and add some tests?

bohdan-struk-avalabs (dismissed):
Code looks good to me. Small comment, not a block.

apps/next/src/pages/Fusion/contexts/hooks/useTransferManager.ts
```diff
@@ -19,6 +25,7 @@ export const useTransferManager = () => {
   const serviceInitializers = useTransferServiceInitializers(
     enabledServices,
     signers,
+    transferStepRef,
```
meeh0w: Syncing this ref through the entire process feels a bit fragile. We needed something similar for recurring swaps and decided to allow a `signerContext` to be passed to [`executeFirstFill`](https://github.com/ava-labs/consumer-sdks/blob/4fc5e601b279a1a79e2832ca59b59f6f62802437/packages/fusion-sdk/src/transfer-service/markr/recurring/types.ts#L338-L344).

What it does is it simply passes it down to the `.sign()` and `.signBatch()` calls.

I think this works a bit better -- there is no temporary dependency that needs maintaining/cleanup in the client app. Every `.transfer()` call would be guaranteed to pass down the proper context that you can later read in the signer implementations.

meeh0w: Actually after looking at it a bit closer, would it not be better to pass the whole `TransferStep` (along with `Quote`) to the `avalancheSendTx()` as the 2nd parameter? Similar to what we do for [`EvmSigner.sign`'s 3rd param](https://github.com/ava-labs/consumer-sdks/blob/19a6a936fab1f3ea3810271f741c57bc9bfcaf5c/packages/fusion-sdk/src/types/signer.ts#L30).

It carries more info as well (total number of required signatures, current signature reason, etc.)

Also IIRC, the `onStepChange` callback was something we wanted to deprecate, since the `TransferStep` was being passed to the signing functions already.

apps/next/src/pages/Fusion/contexts/hooks/useAvalancheFunctions.ts
```diff
@@ -0,0 +1,312 @@
...
+            if (utxoAddresses.has(stripAddressPrefix(address.address))) {
+              indices.push(address.index);
+            }
+
+            return indices;
+          }, [] as number[]);
+
+        const transferStep = transferStepRef?.current;
+        const context =
+          transferStep?.quote.serviceType === ServiceType.AVALANCHE_CCT
```
meeh0w: Can this condition ever be `false`? 🤔

apps/next/src/pages/Fusion/contexts/hooks/useAvalancheFunctions.ts
```diff
@@ -0,0 +1,312 @@
...
+  getUtxos: () => {
+    throw new Error('Avalanche CCT is not available for the active account.');
+  },
+  getWalletAddressesForChainAlias: () => {
+    throw new Error('Avalanche CCT is not available for the active account.');
+  },
+  getWalletChangeAddressForChainAlias: () => {
+    throw new Error('Avalanche CCT is not available for the active account.');
+  },
+};
```
meeh0w, line 72: Something that AI found: 

> `unavailableAvalancheFunctions` throw synchronously. They're typed against callbacks that return `Promise<...>` (`GetAtomicUtxosCallback`, etc.). A `() => { throw }` has return type `never` so it type-checks, but at runtime it throws synchronously rather than rejecting. Any SDK call site doing `.catch()` (rather than `await`) would get an uncaught throw. Minor, but making them `async () => { throw }` (or `() => Promise.reject(...))` matches the declared contract and is safer.

apps/next/src/pages/Fusion/contexts/FusionStateContext.tsx
```diff
@@ -142,11 +149,7 @@ export const FusionStateContextProvider: FC<{ children: ReactNode }> = ({
   const isAmountHigherThanBalance =
     sourceAmountBigInt > (sourceToken?.balance ?? 0n);
 
-  const isAmountLowerThanMinimum =
-    typeof minimumTransferAmount === 'bigint' &&
-    sourceAmountBigInt < minimumTransferAmount;
-
-  const skipFetching = isAmountHigherThanBalance || isAmountLowerThanMinimum;
+  const skipFetching = isAmountHigherThanBalance;
```
meeh0w: Can these checks only be removed for CCT-specific routes? 🤔 

Right now we'll be fetching below-minimum quotes for all services, which increases the traffic, rerenders, etc.

meeh0w (changes requested):
It's looking great overall! I flagged some potential issues, posted some questions.

apps/next/src/pages/Fusion/contexts/hooks/useTransferManager.ts
erictaylor: Good callout. The ref stuff felt gross to me initially. So I agree having the step details passed as an arg on the functions makes way more sense.

I refactored this.

apps/next/src/pages/Fusion/contexts/hooks/useAvalancheFunctions.ts
erictaylor: Refactored.

apps/next/src/pages/Fusion/contexts/hooks/useAvalancheFunctions.ts
erictaylor: Refactored.

apps/next/src/pages/Fusion/contexts/FusionStateContext.tsx
erictaylor: Addressed in f24a50a.

apps/next/src/components/TokenSelect/hooks.ts
erictaylor: Addressed in f24a50a.

meeh0w (dismissed):
Nice 👌 ✅

bohdan-struk-avalabs (approved):

meeh0w (approved):

PR #966, opened 2026-06-30
fix(security): implement SIWE domain mismatch detection

chriselbring-avalabs:
Ported SIWE domain mismatch logic from core-mobile and created an alert

# Summary

Jira ticket: CP-14527
Figma: No figma for this.

I reused the alert from blockaid alerting. Everything else is ported pretty much 1:1 from the implementation in core-mobile.

## Notes

Parses an SIWE signature request and then checks the dapp origin vs the {domain} and {URI} fields https://eips.ethereum.org/EIPS/eip-4361

## Testing

Test to make sure that SIWE mismatches cause an alert and that wellformed SIWE requests do not cause an alert to fire.

## Media
Screenshot 2026-06-22 at 3 56 40 PM
Screenshot 2026-06-22 at 3 56 52 PM

### DIFF
--- a/packages/service-worker/src/vmModules/ApprovalController.test.ts
+++ b/packages/service-worker/src/vmModules/ApprovalController.test.ts
@@ -493,6 +493,84 @@ describe('src/background/vmModules/ApprovalController', () => {
         expect(await promise).toEqual({ signedData: signedTx });
       });
     });
+
+    describe('SIWE domain mismatch detection', () => {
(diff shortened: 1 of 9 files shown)


meeh0w (commented):
Still have to look through it, but I have a feeling this should live in the [EVM Module](https://github.com/ava-labs/core-vm-modules/tree/main/packages/evm-module) instead (there is a handler dedicated to `personalSign` call)

PR #970, opened 2026-07-01
feat: migrate swap tokens v2

bohdan-struk-avalabs:
# Summary

Jira ticket:
https://ava-labs.atlassian.net/browse/CP-14665

https://github.com/user-attachments/assets/c76149c4-a709-47d7-aa27-6ada3c329117

**1. PR Description & .env changes**
  The PR migrates Swap's target token list from the static SDK-bundled list to a live paginated Token Aggregator /v2/tokens
  **API. Key changes:**
  - New service worker handler: GET_MARKR_TARGET_CHAIN_ASSETS — fetches bridgeable tokens per chain with pagination and
  keyword/address search
  - New hook: `useBridgeableTargetTokenList` — infinite scroll, smart pagination, search-by-address-or-keyword
  - `TokenSelect/TokenAmountInput` extended with `onEndReached`, `onOpenChange`, `externalChainOptions`, `isLoadingTokens`, selectedTokenFallback props
  
  **2. Testing preparation & scenarios**
  - Client regeneration — the SDK client (sdk.gen.ts) was regenerated from the Token Aggregator OpenAPI spec. If the spec
  changed, the team needs to re-run the generator and verify the new types match what the backend returns.
  - Scenarios to test:
    - Search by keyword (e.g. "USDC") — paginated results, infinite scroll triggers
    - Chain chip change — token should not deselect while browsing, only commit on select
    - Source token change — target token list should refresh correctly
    - No results state — loading vs. empty vs. error
    

  **3. Do Swap and Send now use different token lists?**
  Yes. They now use completely different sources:
Here is the link to follow up item to update the list in other places 

  `TOKEN_AGGREGATOR_SERVICE_URL`
   These need to be added to .env.dev locally before testing.
   
   Follow up ticket to update token list in other places
   https://ava-labs.atlassian.net/browse/CP-14699


apps/next/src/pages/Fusion/contexts/hooks/useTransferServiceInitializers.ts
```diff
@@ -1,23 +1,40 @@
...
+import { useCallback, useMemo } from 'react';
 import { ServiceType } from '@avalabs/fusion-sdk';
 
-import { useNetworkContext } from '@core/ui';
-import { getServiceInitializer } from '@core/common';
-import { UnifiedTransferSigners } from '@core/types';
+import { useConnectionContext, useNetworkContext } from '@core/ui';
+import { getServiceInitializer, GetTargetChainAssets } from '@core/common';
+import { ExtensionRequest, UnifiedTransferSigners } from '@core/types';
+import { GetMarkrTargetChainAssetsHandler } from '~/services/tokens/handlers/getMarkrTargetChainAssets';
```
meeh0w: 1. This needs an `import type` notation
2. This should be re-exported through `packages/service-worker/src/index.ts` and imported through `@core/service-worker` instead.

meeh0w (changes requested):
Just a quick glance for now, will review deeper a bit later

erictaylor:
We'll want to reconcile changes/diffs in https://github.com/ava-labs/core-extension/pull/965. Also we should maybe try to use the sorting utils @LMulvey introduced in the fusion sdk for sorting.

meeh0w:
@copilot have your comments been properly addressed here?

meeh0w (changes requested):
I started clicking through it first and have some comments before I start reviewing:

1. Please describe the changes in the PR. I see there are some `.env` file adjustments that are needed, but it would be nice to have them documented -- and new env vars to be placed in the proper note in 1Password vault.
2. Please outline the testing preparation and testing scenarios (e.g. client regeneration)
3. Does this mean **Swap** and **Send** (and `Manage tokens` and other places) now use different token lists? If so, is there a follow-up ticket to unify these across the app?

And then from the videos, it seems some of the auto-selection magic is now gone (e.g. `BTC` only has one route and it's to be swapped to `BTC.b` -- it's no longer auto-selected):

https://github.com/user-attachments/assets/33af3a46-03ed-4bc0-bbf6-5df70f212071


I'll work on reviewing the code now. 👀

erictaylor:
`TOKEN_AGGREGATOR_SERVICE_URL` already existed in the ENV. Is there a reason why we added separate PROD and DEV envs? IF we really need both DEV and PROD envs at the same time, we should probably remove the original `TOKEN_AGGREGATOR_SERVICE_URL` and migrate any usage to an appropriate env, but I feel like maybe we don't need these new envs... ?

meeh0w (dismissed):
LGTM. The BTC -> BTC.b autoselection still did not seem to work for me, but I'll let QA team verify :)

meeh0w (dismissed):

erictaylor (commented):
I'm noticing a couple issues when comparing behavior between Core Web and this branch as it relates to the Swap feature.

When selecting a destination asset after selecting AVAX from C/P/X chain as source, the resulting network filter pills don't follow the "C-Chain", "P-Chain", and "X-Chain" name. Likewise, the assets that are populating the list (ie AVAX) appear to be missing their icon logos.

See image:

Image

packages/common/src/unifiedTransfer/assetMapping.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const mapApiTokenToAsset = (
+  token: ApiToken,
+): AssetWithExtras | undefined => {
+  const base = {
+    name: token.name,
+    symbol: token.symbol,
+    decimals: resolveDecimals(token),
+    logoUri: token.logoUri ?? undefined,
+    extra: {
```
erictaylor: This should be `extras`, not `extra`. `AssetWithExtras` exposes provider metadata through `extras`, and the UI reads `asset.extras?.isVerified` when mapping bridgeable tokens and deciding C-Chain pagination. With the current property name, all Markr API tokens lose verification metadata at runtime, so unverified tokens can be displayed as verified and the pagination stop condition treats them as verified as well.

apps/next/src/components/TokenSelect/utils.ts
```diff
@@ -65,15 +65,30 @@ export const areTokenListsEqual = (
...
+
   if (
     prevProps.id !== nextProps.id ||
     prevProps.tokenId !== nextProps.tokenId ||
     prevProps.query !== nextProps.query ||
     prevProps.hint !== nextProps.hint ||
     prevProps.chainFilterMode !== nextProps.chainFilterMode ||
     prevProps.disabled !== nextProps.disabled ||
+    prevProps.selectedChainId !== nextProps.selectedChainId ||
+    prevProps.isLoadingTokens !== nextProps.isLoadingTokens ||
```
erictaylor, line 83: This comparator still looks too narrow for the bridgeable target list. `useBridgeableTargetTokenList` enriches tokens with balances and sorts by balance, but this equality check ultimately compares token-list contents only by ID and `isVerified`. If balances arrive/change while IDs stay the same, `TokenSelect` can skip the render that should update the displayed balances and reorder the menu. Consider including order and the display/sort fields used by token rows, or bypassing this memo equality for the paginated bridgeable list.

erictaylor (approved):

meeh0w (approved):

PR #978, opened 2026-07-09
feat: fall back to RPC calls for unindexed custom tokens

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14667
Figma: _n/a_
Related VM Modules PR: https://github.com/ava-labs/core-vm-modules/pull/442

## TODOs
- [ ] Release the VM Modules PR and update to stable versions here

## Notes
The Core Balance API does not accept `customTokens` as an input parameter balances and therefore this setting is now ignored for chains listed as "supported" by Core Balance API.

We also currently do not check if the balance for user's custom tokens have been returned by the Balance API call, so they may often be completely omitted in the UI.

This PR adds a fallback for custom tokens, such that if balances for them is not returned by the Core Balance API, we will try to query those balances using VM Modules (through direct RPC calls).

## Testing
_coming soon_

## Media
-


apps/next/package.json
```diff
@@ -25,7 +25,7 @@
     "@avalabs/fusion-sdk": "0.26.0",
     "@avalabs/k2-alpine": "1.266.0",
     "@avalabs/types": "3.1.0-alpha.87",
-    "@avalabs/vm-module-types": "3.9.3",
+    "@avalabs/vm-module-types": "0.0.0-fix-custom-tokens-balances-for-20260709075511",
```
erictaylor: We should probably ship with a non canary version right.

packages/inpage/package.json
```diff
@@ -14,8 +14,8 @@
     "typecheck": "yarn tsc --skipLibCheck --noEmit"
   },
   "dependencies": {
-    "@avalabs/evm-module": "3.9.3",
-    "@avalabs/svm-module": "3.9.3",
+    "@avalabs/evm-module": "0.0.0-fix-custom-tokens-balances-for-20260709075511",
+    "@avalabs/svm-module": "0.0.0-fix-custom-tokens-balances-for-20260709075511",
```
erictaylor: Same comment here. We should maybe update to a stable version and not ship canary versions.

bohdan-struk-avalabs (approved):
Code looks good!

erictaylor (approved):

PR #979, opened 2026-07-09
fix: hide gasless switch for batch approvals

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14719
Figma: _n/a_

## Notes
* Hides the gasless switch on batch-approval screens, where it would only ever fund the first transaction in the batch, making the whole batch fail for people with not enough AVAX.

## Testing
0. Enable quick swaps in the settings.
1. Perform a simple send on C-Chain - gasless switch should be visible and working.
2. Perform a 2-step swap on C-Chain (e.g. approve + transfer) - gasless switch should not be visible.


erictaylor (approved):

bohdan-struk-avalabs (approved):
Tested, looks good!

PR #982, opened 2026-07-13
fix: avalanche cct qa fixes

erictaylor:
This PR addresses a couple small things caught by QA during review that need fixed before shipping.

1. "Select UTXOs" button gets squashed because of other inline button sizes.
2. "This swap will require two transactions" warning message for avalanche cct swaps text is to large and needs sized down
3. Avalanche CCT swap notification messages need more detail.
4. Message for stuck atomics recovery can display immediately after successful swap.

Fixes:
1. Decreased horizontal button padding for 25%, 50%, and Max buttons. All buttons now fit in default extension width.
2. Message text sized down to "caption" style.
3. Notification message for Avalanche CCT transfers now are formatted like: "Swapped AVAX (C-Chain) to AVAX (P-Chain)".
4. Adjusts logic for stuck atomics message display. Does not display if there are any pending avalanche cct transfers. Also waits for a grace period after completed transfers for balance service to refresh before displaying message.

### DIFF
--- a/apps/next/src/components/TokenAmountInput/TokenAmountInput.tsx
+++ b/apps/next/src/components/TokenAmountInput/TokenAmountInput.tsx
@@ -1,5 +1,6 @@
 import { TokenUnit } from '@avalabs/core-utils-sdk';
 import {
+  ButtonProps,
   CircularProgress,
   Collapse,
   Grow,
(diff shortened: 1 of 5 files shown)


erictaylor:
@copilot are all your previous review comments addressed?

meeh0w (approved):

bskyn (approved):

PR #988, opened 2026-07-17
feat: enable hyperevm swap

bskyn:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14552
Dependent on: https://github.com/ava-labs/consumer-sdks/pull/294

## Notes
- Add a signTypedData implementation to the Fusion EVM signer, routing typed-data requests through eth_signTypedData_v4 so the SDK can sign the Hyperliquid evmUserModify L1 action that enables HyperEVM big blocks
- Reconstruct the required EIP712Domain type entry from the domain fields present (the SDK's payload omits it), and sign the payload's own domain.chainId verbatim so the Hyperliquid L1 pseudo-chain (1337) is preserved rather than overridden by the routing chainId.
- Pass the routing chainId only to select the signing network/account, keeping message construction untouched
- Copy clean-up

## Testing
- Load extension locally
- Select HyperEVM on swap
- Test HYPE -> USDC and verify

## Media
For swaps that have large gas estimate, this message will be shown so that user can enable big blocks on HyperEVM during swap.
Screenshot 2026-07-17 at 2 50 34 PM

Without approving this, we get internal error for gas estimation in core-extension

Video with successful swap:
https://github.com/user-attachments/assets/0be872ec-a741-4918-b202-99e9801ca23a


apps/next/src/pages/Fusion/lib/signers/evm.ts
```diff
@@ -30,6 +30,22 @@ const getChainIdForBatch = (transactions: readonly EvmTransactionRequest[]) => {
...
+// eth_signTypedData_v4 (and the eth-sig-util signer it ultimately routes to)
+// require an explicit `EIP712Domain` type entry, which the SDK's typed-data
+// payload omits. Rebuild it from whichever domain fields are actually present.
+const EIP712_DOMAIN_FIELD_TYPES: Record<string, string> = {
+  name: 'string',
+  version: 'string',
+  chainId: 'uint256',
+  verifyingContract: 'address',
+  salt: 'bytes32',
+};
```
meeh0w: My immediate question: why does the SDK omit it? Wouldn't it be better for it to be supplied by the SDK, so the client apps don't need to recreate the `EIP712Domain` field on its own?

If we do have to recreate it in the client apps, can we use `getTypesForEIP712Domain()` from `viem`?

meeh0w (changes requested):
Just one blocking question/comment; other than that LGTM.

apps/next/src/pages/Fusion/lib/signers/evm.ts
bskyn: Comment here in the sdk says it mirrors viem shape which is why its omitted
https://github.com/ava-labs/consumer-sdks/blob/735626da06b9cd3b96870e760fa02e8a871cdaea/packages/fusion-sdk/src/types/signer.ts#L42-L61

Adding it in the SDK would:
- Fight the viem-compat contract
- Touch every builder (buildL1ActionAgentTypedData, Relay/SendAsset, Markr pass-through, etc.)
- Still be redundant for wagmi clients (viem rebuilds it anyway)

Updated the PR to use viem’s `getTypesForEIP712Domain` https://github.com/ava-labs/core-extension/pull/988/commits/c161636ac1ad7e5c354dd4e37ae510315d3a1d5f

meeh0w (dismissed):

erictaylor (dismissed):

erictaylor (approved):

meeh0w (approved):

PR #993, opened 2026-07-22
feat: recognize encrypted erc operations

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14801
Figma: https://www.figma.com/design/kksuPP4PnRIFDx9Sm3544F/Core-Extension-2026?node-id=12910-67060&t=HMPmwnVYdsodwHxF-0

## Notes
1. Updates VM Modules to the latest version (EVM Module @ 3.13.0 introduces support for recognizing eERC20 transactions)
2. Updates `ModuleManager` to supply the new `features.encryptedERCs` runtime parameter to the `EVMModule` constructor (and update it as feature flags change).

## Testing
1. Go to https://www.encryptederc.com/ and connect Core
2. When completing the registration, you should see the `Privacy: eERC 20` and `Operation: register` line items on the approval screen.
3. When performing any other operations (transfer / deposit / withdraw) you should see `Privacy: eERC 20` and the name of the operation on the approval screens.
4. On `deposit` / `withdraw` you'll see the non-encrypted tokens leaving/entering your wallet, but _*not*_ their encrypted equivalents (e.g. you'll see `-1 USDC` on deposit, but not the `+1 eUSDC`).
5. On `transfer` operation, you will not see the balance change at all.

## Media
image
image
image


erictaylor (approved):
Looks good

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+      measureDuration(txHash).start();
+
+      return {
+        name: 'eth_sendTransaction_success',
+        properties: {
+          dAppUrl: request.dappInfo.url,
+          address: request.params[0].from,
+          txHash,
+          chainId: request.chainId,
+          privacy: getPrivacyData(request),
```
B0Y3R-AVA, line 41: Is it intentional that this isn't gated by `eerc-transaction-display`? The module only parses eERC data for the approval screen when the flag is on, but this analytics path ships to everyone the moment we merge, with no kill switch. And since `parseEercTransaction` matches by selector against any contract (the SDK comment itself calls it "a decode hint, not a trust signal"), we'll tag unrelated contracts as `eerc20` in PostHog whenever selectors collide. If the goal is measuring adoption before turning the display on, that works, but I want to confirm it's a decision and not an oversight.

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      measureDuration(txHash).start();
+
+      return {
+        name: 'eth_sendTransaction_success',
+        properties: {
+          dAppUrl: request.dappInfo.url,
+          address: request.params[0].from,
+          txHash,
+          chainId: request.chainId,
```
B0Y3R-AVA, line 40: Heads up: `request.chainId` here is the CAIP-2 string (`eip155:43114`), but the avalanche handlers next door send the numeric `network?.chainId`, and `updateChainIdIfNeeded` in AnalyticsServicePosthog only normalizes numbers. Dashboards filtering `chainId = 43114` will match the avalanche events and miss these. Probably worth resolving the network and reporting the numeric id like the sibling handler does.

packages/inpage/package.json
```diff
@@ -14,8 +14,8 @@
     "typecheck": "yarn tsc --skipLibCheck --noEmit"
   },
   "dependencies": {
-    "@avalabs/evm-module": "3.10.0",
-    "@avalabs/svm-module": "3.10.0",
+    "@avalabs/evm-module": "3.13.0",
```
B0Y3R-AVA: This 3.13.0 never actually installs. Root resolutions pin evm-module and svm-module to 4.0.1, and yarn.lock has no 3.13.0 entry at all. Harmless today (I diffed the `dist/provider` files inpage imports between the two tags, byte-identical), but the manifest is lying about what ships in the injected script, and it becomes a trap if anyone removes the resolutions entry thinking it's redundant. I'd declare 4.0.1 here.

bohdan-struk-avalabs (dismissed):

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
meeh0w: I say we want to measure how many eERC transactions are signed through Core, not necessarily how many times we display that information based on the feature flag.

As for `decode hint, not a trust signal` - this is a limitation we're aware of.

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
meeh0w: We intended to move away from numeric chain IDs towards CAIP-2 identifiers (started with Fusion integration, I believe).

Also, `eth_sendTransaction_*` events have not been collected for _a long time_ now in Core Extension (ever since we moved from internal `eth_sendTransaction.ts` handler to the EVM Module handling), so with this we're basically doing a clean slate.

packages/inpage/package.json
meeh0w: Good catch, not entirely sure how I managed to update these to the wrong version instead of `4.0.1`. Probably a bad conflict resolution. Fixed in f00663e99

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+    request.method === RpcMethod.ETH_SEND_TRANSACTION &&
+    Array.isArray(request.params) &&
+    request.params.length === 1
+  );
+};
+
+const getPrivacyData = (
+  request: EthSendTransactionRequest,
+): PrivacyData | undefined => {
+  const { operation } = parseEercTransaction(request.params[0]) ?? {};
```
gergelylovas, line 101: Can this parsing happen in the vm-modules?

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
```diff
@@ -0,0 +1,103 @@
...
+        },
+      };
+    },
+    [TransactionStatusEvents.CONFIRMED]: async (event) => {
+      if (!isEthSendTransactionRequest(event.value.request)) {
+        return null;
+      }
+
+      const { txHash, request } = event.value;
+      const duration = measureDuration(txHash).end();
```
B0Y3R-AVA, line 71: `measureDuration().end()` calls `performance.measure(name, '${txHash}-start')`, which throws a DOMException when the start mark doesn't exist. In MV3 the service worker is torn down after ~30s idle and its `performance` marks die with it, but an `eth_sendTransaction` usually confirms well after that, in a fresh SW where the `PENDING` start mark is gone.

When that happens here (and in the `REVERTED` branch above), `.end()` throws, the dispatcher awaits the handler with no try/catch (`TransactionStatusEventsSubscriber.#onTransactionStatusEvent`), so the promise rejects unhandled and the `_confirmed`/`_failed` event is never captured. Because `eth_sendTransaction` is high-volume this will actually fire in the field.

This mirrors the existing `avalancheSendTransactionHandler` pattern, so it's a shared latent issue rather than something this PR introduces alone — but it's worth guarding `.end()` (try/catch → `duration: undefined`, or verify the mark exists), ideally inside `measureDuration` so both handlers benefit.

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
B0Y3R-AVA, line 101: `getPrivacyData` calls `parseEercTransaction` unconditionally, so every `eth_sendTransaction_*` analytics event is tagged `privacy: { type: 'eerc20', operation }` regardless of `EERC_TRANSACTION_DISPLAY`. The approval-screen display, by contrast, is gated (ModuleManager only sets `features.encryptedERCs` from the flag). So with the flag OFF (its default), users won't see any eERC UI but the eERC classification still ships in telemetry.

If that recognize-always / display-gated split is intentional, all good — just flagging since both the ticket and the flag name are framed around "display," so wanted to confirm the analytics side is meant to ignore the gate.

meeh0w: It does happen in the VM Modules for the approval screen, this part here is solely for analytics (using the util function exported from EVM Module).

I did not want to add another flag to `DisplayData` just for the sake of analytics, and did not want to make  analytics dependent on literal strings in the `DisplayData` either (i.e. looking for `Privacy` line item) as this would be fragile (and not give us any analytics on eERC usage unless the feature flag is on).

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
meeh0w: [We keep the service worker alive](https://github.com/ava-labs/core-extension/pull/993#discussion_r3707326975), but the browser could be restarted in the meantime. Added a `try..catch` here: a5ee9e348

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
meeh0w: https://github.com/ava-labs/core-extension/pull/993#discussion_r3680680482

bohdan-struk-avalabs (approved):

B0Y3R-AVA (approved):

packages/service-worker/src/services/transactions/handlers/ethSendTransactionHandler.ts
meeh0w: @gergelylovas I'm merging this for now, let me know if you'd like to change the approach here; happy to discuss.

gergelylovas: Ack. I've been thinking about this DisplayData shortcoming related to the MCP server as well. We probably should introduce some IDs or similar for them in the future so they can be more relied on.

PR #999, opened 2026-07-27
chore: migrate to Core Glacier Proxy

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14867
Figma: _n/a_

## TODOs
- [x] Add Core Proxy Glacier URL & Core Proxy API key in secrets / 1Pass ❗ ⚠️ 

## Notes
1. Implements `getAuthHeaders()` for EVM Module and Avalanche Module
2. Removes the usage of `process.env.GLACIER_API_KEY` as the proxy rate-limiting key for `BitcoinProvider`
3. Removes `addGlacierApiKeyIfNeeded` helper and its usages
4. Cleans up the `GlacierService` from unused methods and makes sure the built client has the `X-Firebase-Appcheck` header appended.

## Testing
No functionality change -- regular regression/sanity check.


bohdan-struk-avalabs (dismissed):

packages/service-worker/src/vmModules/ModuleManager.ts
```diff
@@ -76,21 +83,39 @@ export class ModuleManager {
...
+          ? { 'X-Firebase-AppCheck': appcheckToken.token }
+          : {};
+      } catch (error) {
+        console.error('Error getting AppCheck token:', error);
+        Monitoring.sentryCaptureException(
+          error as Error,
+          Monitoring.SentryExceptionTypes.FIREBASE,
+        );
+
+        return {};
```
B0Y3R-AVA: What happens when AppCheck is down or just hasn't finished attesting yet? This returns `{}`, the request goes out bare, and the new proxy rejects it. On 4.0.0 that includes the Glacier UTXO fetch inside `avalanche-sign-transaction`, so X/P sign/send would break for as long as AppCheck is unhappy. Web and mobile both keep an `x-core-api-key` fallback for exactly this case — was skipping it here deliberate?

packages/service-worker/src/vmModules/ModuleManager.ts
```diff
@@ -76,21 +83,39 @@ export class ModuleManager {
...
+      }
+    };
+
     this.#modules = [
       new EvmModule({
         environment,
         approvalController: this.#approvalController,
         appInfo,
-        runtime: { fetch: circuitBreakerFetch },
+        runtime: { fetch: circuitBreakerFetch, getAuthHeaders },
```
B0Y3R-AVA, line 85: One ordering thing I noticed: `BackgroundRuntime` calls `moduleManager.activate()` before `appCheckService.activate()`, and `EvmGlacierService` fires a `supportedChains` probe straight from its constructor. So every MV3 cold start sends one unauthenticated probe before AppCheck exists. It's noop-caught and heals on the next call, but service workers cold start constantly — worth swapping the two activate calls?

packages/service-worker/src/services/glacier/GlacierService.ts
```diff
@@ -1,80 +1,42 @@
...
-
-      // If we reached max. attempts and NFT is still not updated,
-      // throw a recognizable error.
-      if (shouldPoll && fetchCount >= maxAttempts) {
-        throw CommonError.RequestTimeout;
+  async #getClient() {
+    try {
+      const appcheckToken = await this.appCheckService.getAppcheckToken();
+      if (!appcheckToken) {
+        throw new Error('No appcheck token found');
```
B0Y3R-AVA: Here we throw when the token's missing, but `getAuthHeaders` in ModuleManager swallows the same failure and returns `{}` — which behavior do we actually want? Asking because the only caller left is `getUnknownUsedNetwork`, which AccountsService fires without await or catch, so this throw lands as an unhandled rejection in the SW and the used-network auto-enable just silently doesn't happen. It also looks like it double-reports to Sentry (the capture in the catch here plus the global handler).

packages/service-worker/src/services/glacier/GlacierService.ts
```diff
@@ -1,80 +1,42 @@
...
       }
-    } while (shouldPoll);
-
-    return token;
+      return new Glacier({
+        BASE: process.env.GLACIER_URL,
```
B0Y3R-AVA: This still points at `GLACIER_URL`, so we're fully coupled to the secret flip in the TODO. `listAddressChains` is one of the EVM endpoints the old host is losing, so if a build ships before the secret changes, unknown-network detection quietly breaks. Do we have the release ordering nailed down with infra — same release for both?

packages/common/src/utils/network/getProviderForNetwork.ts
```diff
@@ -41,14 +39,9 @@ export const getProviderForNetwork = async (
       `${process.env.PROXY_URL}/proxy/nownodes/${
         network.isTestnet ? 'btc-testnet' : 'btc'
       }`,
-      process.env.GLACIER_API_KEY
-        ? { rltoken: process.env.GLACIER_API_KEY }
-        : {},
     );
```
B0Y3R-AVA, line 47: Removing this makes sense for the repo-local providers, but bitcoin-module and svm-module at 4.0.0 still read `process.env.GLACIER_API_KEY` themselves (`get-provider.ts` in both), and `GlacierFetchHttpRequest` still appends `rltoken` for the evm/avalanche Glacier clients. So the lavamoat grants for those modules are still load-bearing, and if we delete the secret as part of this migration, BTC and Solana silently lose the rate-limit bypass. What's the plan there — keep the secret alive until vm-modules drops those reads?

packages/service-worker/src/vmModules/ModuleManager.ts
meeh0w: My understanding is `X-Core-Api-Key` should only be used for QA tests (which can trigger rate limiters) and for SSR on Core Web.

I don't think it's a good idea to place it in our client bundles - we'd be exposing the key to the public at this point.

WDYT @gergelylovas ?

packages/common/src/utils/network/getProviderForNetwork.ts
meeh0w: I do not intend to remove the secret altogether, just stopped using it here...although maybe I was too fast and it should stay. I'll revert it.


BTW this secret is not supposed to be bundled in the production builds, so users would not lose any functionality here:

https://github.com/ava-labs/core-extension/blob/d11d47480714ab37d0f87de209567df3d0fb241a/.env.example#L21-L24

packages/service-worker/src/services/glacier/GlacierService.ts
meeh0w: Added a specific `CORE_PROXY_GLACIER_BASE_URL` env var.

packages/common/src/utils/jsonRpcEngine.ts
```diff
@@ -1,12 +1,11 @@
 import { JsonRpcEngine } from 'json-rpc-engine';
 import { createFetchMiddleware } from 'eth-json-rpc-middleware';
 import { Network } from '@avalabs/core-chains-sdk';
-import { addGlacierAPIKeyIfNeeded } from './network/addGlacierAPIKeyIfNeeded';
 
 export async function engine(network: Network) {
   const fetchMiddleware = createFetchMiddleware({
     get rpcUrl() {
-      return addGlacierAPIKeyIfNeeded(network.rpcUrl);
+      return network.rpcUrl;
```
gergelylovas: Since the RPCs won't go behind the glacier proxy, we might still need rate limit bypass in here for e2e test environments.

packages/common/src/utils/network/getProviderForNetwork.ts
```diff
@@ -46,9 +44,7 @@ export const getProviderForNetwork = async (
         : {},
     );
   } else if (network.vmName === NetworkVMType.EVM) {
-    const fetchConfig = new FetchRequest(
-      addGlacierAPIKeyIfNeeded(network.rpcUrl),
-    );
+    const fetchConfig = new FetchRequest(network.rpcUrl);
```
gergelylovas: Same here regarding rate limiting on the RPCs

packages/service-worker/src/vmModules/ModuleManager.ts
B0Y3R-AVA: Fair enough on not shipping the key in client bundles — agreed for prod. Two things in the follow-up commit seem to cut against that though. `create_release.yaml` now writes `CORE_PROXY_API_KEY` into `.env.production`, so we're one provisioned secret away from baking it into a public bundle — should that live only in the dev/e2e workflows? And `getAuthHeaders` sends `X-Core-Api-Key: ''` whenever the env is unset, which is every prod request. If the proxy validates the header whenever it's present, wouldn't an empty value fail auth even with a good AppCheck token? Conditionally spreading it like the AppCheck header would dodge both. @gergelylovas do you know how the proxy treats an empty `x-core-api-key`?

packages/service-worker/src/vmModules/ModuleManager.ts
B0Y3R-AVA: The activate() swap covers it, thanks. 👍

packages/service-worker/src/services/glacier/GlacierService.ts
B0Y3R-AVA: Much cleaner — shared resolver, no throw, single Sentry capture. Thanks.

packages/service-worker/src/services/glacier/GlacierService.ts
B0Y3R-AVA: The dedicated env var is a better shape, thanks. One thing before merge though: if `CORE_PROXY_GLACIER_BASE_URL` isn't set, `BASE` comes through undefined and the SDK quietly falls back to its baked-in default, which is the old glacier host. Are the new secrets already provisioned in the repo, or is that still part of the TODO?

packages/common/src/utils/network/getProviderForNetwork.ts
B0Y3R-AVA: You're right — I went back and checked `create_release` and `GLACIER_API_KEY` never made it into prod builds, so this was dev/e2e impact only. My bad on the framing. Thanks for restoring it anyway.

packages/service-worker/src/services/appcheck/utils/getAuthHeaders.ts
```diff
@@ -0,0 +1,35 @@
...
+    ? {
+        [FIREBASE_APP_CHECK_HEADER]: appcheckToken,
+        [CORE_PROXY_API_KEY_HEADER]: process.env.CORE_PROXY_API_KEY ?? '',
+      }
+    : {
+        [CORE_PROXY_API_KEY_HEADER]: process.env.CORE_PROXY_API_KEY ?? '',
+      };
+};
+
+const _getAppcheckToken = async (): Promise<string | undefined> => {
```
gergelylovas, line 29: nit: If this were a private method in GlacierService, dependency injection could inject the AppCheckService making dependencies more explicit and easy to detect.

packages/ui/src/contexts/BalancesProvider/BalancesProvider.tsx
```diff
@@ -391,23 +384,6 @@ export function BalancesProvider({ children }: PropsWithChildren) {
     [network, request],
   );
 
-  const refreshNftMetadata = useCallback(
```
gergelylovas, line 394: Did we lose this functionality during the redesign? 🤔

gergelylovas (changes requested):

packages/ui/src/contexts/BalancesProvider/BalancesProvider.tsx
meeh0w: The NFT details page was moved to Core Web :)

packages/common/src/utils/jsonRpcEngine.ts
meeh0w: Oh, true! Restoring...

packages/service-worker/src/services/appcheck/utils/getAuthHeaders.ts
meeh0w: It's also used by the `ModuleManager`, is why I opted for a util function instead of adding it within the `GlacierService` (I did that initially, but then went the other way).

packages/common/src/utils/jsonRpcEngine.ts
meeh0w: Restored.

packages/service-worker/src/vmModules/ModuleManager.ts
meeh0w: Agreed on the `''` fallback - fixed.
Also agreed on not adding an unnecessary write to the env file for production workflow (fixed too).

packages/service-worker/src/services/glacier/GlacierService.ts
meeh0w: Env vars are now added in the repo.

packages/service-worker/src/vmModules/ModuleManager.ts
gergelylovas: Empty string is fine. The backend fallbacks to '' if the key is not defined and has an `if(providedApiKey)` check. Then checks the presence of the appcheck token

packages/ui/src/contexts/BalancesProvider/BalancesProvider.tsx
gergelylovas: ahh true true

gergelylovas (approved):

B0Y3R-AVA (approved):
Verified all my comments were addressed — LGTM. Thanks for the quick turnaround!

bohdan-struk-avalabs (approved):

PR #1008, opened 2026-08-07
refactor: get network enablement state from API

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14920
Figma: n/a

## Notes
* Makes use of the `isAlwaysEnabled` and `isEnabledByDefault` flags returned by the new `/tokenlist` endpoint.
* Still leaves the hardcoded list of always-enabled chains, but when API response comes in, it will take the priority.

## Testing
* Make sure the networks returned as `isAlwaysEnabled` from `/tokenlist` are in fact enabled (and disabling them is not possible)
* Make sure the networks returned as `IsEnabledByDefault` from `/tokenlist` are in fact enabled (but the users can still disable them and their choice takes priority)
* Make sure the networks enabled-state retains its functionality.

### DIFF
--- a/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleList.tsx
+++ b/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleList.tsx
@@ -14,7 +14,7 @@ type NetworkToggleListProps = {
   networks: NetworkWithCaipId[];
 };
 
-const defaultNetworkSet = new Set(NETWORKS_ENABLED_FOREVER);
+const alwaysEnabledNetworkSet = new Set(NETWORKS_ENABLED_FOREVER);
 export const NetworkToggleList = ({ networks }: NetworkToggleListProps) => {
   const { enabledNetworks, enableNetwork, disableNetwork } =
     useNetworkContext();
@@ -104,7 +104,10 @@ export const NetworkToggleList = ({ networks }: NetworkToggleListProps) => {
           key={network.chainId}
           network={network}
           isEnabled={enabledNetworksArray.includes(network.chainId)}
-          isDefault={defaultNetworkSet.has(network.chainId)}
+          isAlwaysEnabled={
+            Boolean(network.isAlwaysEnabled) ||
+            alwaysEnabledNetworkSet.has(network.chainId)
+          }
           onToggle={() => {
             if (enabledNetworksArray.includes(network.chainId)) {
               disableNetwork(network.chainId);
--- a/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleList.tsx
+++ b/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleList.tsx
@@ -1,5 +1,5 @@
 import { Stack } from '@avalabs/k2-alpine';
-import { NETWORKS_ENABLED_FOREVER, NetworkWithCaipId } from '@core/types';
+import { NetworkWithCaipId } from '@core/types';
 import {
   useAccountsContext,
   useBalancesContext,
@@ -14,7 +14,6 @@ type NetworkToggleListProps = {
   networks: NetworkWithCaipId[];
 };
 
-const alwaysEnabledNetworkSet = new Set(NETWORKS_ENABLED_FOREVER);
 export const NetworkToggleList = ({ networks }: NetworkToggleListProps) => {
   const { enabledNetworks, enableNetwork, disableNetwork } =
     useNetworkContext();
@@ -104,10 +103,7 @@ export const NetworkToggleList = ({ networks }: NetworkToggleListProps) => {
           key={network.chainId}
           network={network}
           isEnabled={enabledNetworksArray.includes(network.chainId)}
-          isAlwaysEnabled={
-            Boolean(network.isAlwaysEnabled) ||
-            alwaysEnabledNetworkSet.has(network.chainId)
-          }
+          isAlwaysEnabled={Boolean(network.isAlwaysEnabled)}
           onToggle={() => {
             if (enabledNetworksArray.includes(network.chainId)) {
               disableNetwork(network.chainId);
--- a/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleListItem.tsx
+++ b/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleListItem.tsx
@@ -19,7 +19,7 @@ type NetworkToggleListItemProps = {
 export const NetworkToggleListItem = ({
   network,
   isEnabled,
-  isAlwaysEnabled: isDefault,
+  isAlwaysEnabled,
   onToggle,
   onClick,
 }: NetworkToggleListItemProps) => {
@@ -42,12 +42,12 @@ export const NetworkToggleListItem = ({
         />
       </ListItemIcon>
       <ListItemText primary={network.chainName} />
-      {isDefault ? null : (
+      {isAlwaysEnabled ? null : (
         <Switch
           data-testid={`network-toggle-${network.chainId}`}
           size="small"
           checked={isEnabled}
-          disabled={isDefault}
+          disabled={isAlwaysEnabled}
           onChange={onToggle}
           onClick={(e) => e.stopPropagation()}
         />
--- a/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleListItem.tsx
+++ b/apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleListItem.tsx
@@ -11,7 +11,7 @@ import { NetworkAvatar } from '../NetworkAvatar/NetworkAvatar';
 type NetworkToggleListItemProps = {
   network: NetworkWithCaipId;
   isEnabled: boolean;
-  isDefault: boolean;
+  isAlwaysEnabled: boolean;
   onToggle: () => void;
   onClick: () => void;
 };
@@ -19,7 +19,7 @@ type NetworkToggleListItemProps = {
 export const NetworkToggleListItem = ({
   network,
   isEnabled,
-  isDefault,
+  isAlwaysEnabled: isDefault,
   onToggle,
   onClick,
 }: NetworkToggleListItemProps) => {
--- a/packages/common/src/utils/network/getNetworkEnablement.ts
+++ b/packages/common/src/utils/network/getNetworkEnablement.ts
@@ -6,9 +6,11 @@ import {
 
 /**
  * Hybrid resolution (always-enabled networks):
- *   - the hardcoded `NETWORKS_ENABLED_FOREVER` is the base set
- *   - the API's `isAlwaysEnabled` flag overrides the base set: `true` adds a
- *     network, `false` removes it, `undefined` leaves the base set untouched
+ *   - the hardcoded `NETWORKS_ENABLED_FOREVER` is a floor that the API cannot
+ *     remove from. `/tokenlist` always sends `isAlwaysEnabled` as a boolean
+ *     (never `undefined`); untagged chains come back `false`, so honoring
+ *     `false` would make C-Chain / ETH user-disableable after a Contentful miss.
+ *   - `true` adds a network; `false` / `undefined` leave the floor untouched.
  *
  * Hybrid resolution (default-enabled networks):
  *   - the hardcoded `NETWORKS_ENABLED_BY_DEFAULT` is the base set
@@ -23,8 +25,6 @@ export function getAlwaysEnabledNetworkIds(chainList: ChainList): number[] {
   for (const network of Object.values(chainList)) {
     if (network.isAlwaysEnabled === true) {
       ids.add(network.chainId);
-    } else if (network.isAlwaysEnabled === false) {
-      ids.delete(network.chainId);
     }
   }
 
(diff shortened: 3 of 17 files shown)


apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleList.tsx
```diff
@@ -104,7 +104,10 @@ export const NetworkToggleList = ({ networks }: NetworkToggleListProps) => {
           key={network.chainId}
           network={network}
           isEnabled={enabledNetworksArray.includes(network.chainId)}
-          isDefault={defaultNetworkSet.has(network.chainId)}
+          isAlwaysEnabled={
+            Boolean(network.isAlwaysEnabled) ||
+            alwaysEnabledNetworkSet.has(network.chainId)
+          }
```
B0Y3R-AVA, line 110: This reimplements the always-enabled resolution inline instead of reusing `getAlwaysEnabledNetworkIds()`, and the semantics diverge: `Boolean(network.isAlwaysEnabled) || alwaysEnabledNetworkSet.has(network.chainId)` can never go back to `false` for a chain already in `NETWORKS_ENABLED_FOREVER`, even if the API returns `isAlwaysEnabled: false` for it (`false || true` is still `true`).

`NetworkToggleListItem.tsx` renders `{isDefault ? null : <Switch .../>}`, so when this stays stuck `true` the switch never renders at all — the user has no way to disable that network in Settings, even though `NetworkService`'s `#alwaysEnabledNetworks` (built via the correct utility) would honor the override and allow it. Suggest importing/using `getAlwaysEnabledNetworkIds` here instead, and adding a test on this component for the `isAlwaysEnabled: false` override case.

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -276,14 +288,23 @@ export class NetworkService implements OnLock, OnStorageReady {
...
+
   public get customNetworks() {
     return this._customNetworks;
   }
 
   async enableNetwork(chainId: number) {
     if (
       this._enabledNetworks.includes(chainId) ||
-      NETWORKS_ENABLED_FOREVER.includes(chainId)
+      this.#alwaysEnabledNetworks.includes(chainId)
```
B0Y3R-AVA, line 307: `enableNetwork` now guards on `this.#alwaysEnabledNetworks.includes(chainId)`, but `disableNetwork` (a few lines below) has no equivalent guard — it unconditionally persists `{isEnabled: false}` via `networkAvailability`/`updateNetworkState()`. Two reachable call sites hit this with no protection: `DisableNetworkHandler` (RPC) and the Concierge AI assistant's `disableNetwork` function in `useFunctions.ts`.

Today it's masked at runtime because the `enabledNetworks` setter re-unions `#alwaysEnabledNetworks` back in every time, so nothing looks broken in-session. But it persists a stale `isEnabled: false` record in storage for a chain that's *currently* always-enabled, and since "always-enabled" is now API-controlled and can change over time (rather than a small fixed list), if the API later stops flagging that chain, the stale record silently takes effect. Worth mirroring the same guard in `disableNetwork`.

B0Y3R-AVA (commented):
Adversarial pass over the enablement logic — left two inline notes, both about the new `isAlwaysEnabled`/`isEnabledByDefault` hybrid resolution diverging between call sites. The utility functions in `getNetworkEnablement.ts` themselves look correct.

apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleList.tsx
meeh0w: Fixed in [f1a2f8a](https://github.com/ava-labs/core-extension/pull/1008/commits/f1a2f8ad52e1bd0819bf49acca1937cb1edb1145)

packages/service-worker/src/services/network/NetworkService.ts
meeh0w: Fixed in [f1a2f8a](https://github.com/ava-labs/core-extension/pull/1008/commits/f1a2f8ad52e1bd0819bf49acca1937cb1edb1145)

packages/common/src/utils/network/getNetworkEnablement.ts
```diff
@@ -0,0 +1,50 @@
...
+ *   - always-enabled networks are always included, regardless of the flag
+ */
+
+export function getAlwaysEnabledNetworkIds(chainList: ChainList): number[] {
+  const ids = new Set<number>(NETWORKS_ENABLED_FOREVER);
+
+  for (const network of Object.values(chainList)) {
+    if (network.isAlwaysEnabled === true) {
+      ids.add(network.chainId);
+    } else if (network.isAlwaysEnabled === false) {
```
B0Y3R-AVA: Does this `false` branch ever fire the way the comment above describes? I traced the payload and I don't think `undefined` ever reaches us, which would mean there's no floor here at all.

What I found:

- `getChainsAndTokens` in core-chains-sdk is a raw passthrough (`new HttpClient(url).get('')`). No field mapping, so whatever the API sends lands in the ChainList as-is.
- `${PROXY_URL}/tokenlist` is proxy-api-worker's `handleTokenList`, which spreads `...chain` straight off the aggregator's `/v1/networks`.
- The aggregator declares both as required booleans: `isAlwaysEnabled: z.boolean()` and `isEnabledByDefault: z.boolean()` in `networkListSchema.ts`.
- The cron sets `is_always_enabled` absolutely on every run from Contentful (`p.caip2 = ANY(alwaysEnabledCaip2Ids)`), with no leave-unchanged branch. Anything not tagged `Always enabled` comes back `false`.

If that's right, the API decides both sets outright and `NETWORKS_ENABLED_FOREVER` stops mattering. The chains I'd worry about are the ones `_initChainList` doesn't hardcode: C-Chain 43114, Fuji 43113, ETH 1, Sepolia, and Solana mainnet plus devnet. BTC, P/X, HyperEVM and HyperCore get merged in from SDK constants, so they stay flagless and keep the old behavior. The testnets feel most exposed, since I'd guess the Contentful tagging covers mainnets only.

Do you know what Contentful says today for those six? If C-Chain isn't tagged it becomes user-disableable, and we lean on it as the fallback in `getAvalancheNetwork`, `uiActiveNetwork` and `getInitialNetworkForDapp`.

Would it be worth making the API additive only here, so keep the `true` branch and drop the `false` delete? A Contentful edit turning off C-Chain is a bad failure mode to leave open. And maybe a test where the API sends `isAlwaysEnabled: false` for 43114, since every test right now goes down the `undefined` path that I don't think prod ever takes.

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -276,14 +289,23 @@ export class NetworkService implements OnLock, OnStorageReady {
     return this.#filterByEnvironment(this._enabledNetworks);
   }
 
+  /**
+   * Networks that should be enabled by default on onboarding. Derived from the
+   * chainlist (API `isEnabledByDefault` flag) with the hardcoded list as a floor.
+   */
+  async getNetworksEnabledByDefault(): Promise<number[]> {
```
B0Y3R-AVA, line 292: `allNetworks` doesn't run `#filterBasedOnFeatureFlags` or `#filterBasedOnDevMode`, only `activeNetworks` does (line 172). So if the API marks a gated network `isEnabledByDefault: true`, does onboarding enable it and persist `isEnabled: true` for something we're currently hiding? It would then read as already-on the moment someone flips the gate. We hit a version of this on mobile when a BE flag flip leaked HyperEVM into the L1 list.

Worth gating on the feature flags here? I don't think swapping in `activeNetworks` is the fix, since the dev-mode filter would drop the testnet ids both constants deliberately carry.

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -932,6 +959,36 @@ export class NetworkService implements OnLock, OnStorageReady {
...
+      ]),
+    );
+  };
+
+  #updateEnablementFromChainList = (chainList?: ChainList) => {
+    if (!chainList) {
+      return;
+    }
+
+    this.#alwaysEnabledNetworks = getAlwaysEnabledNetworkIds(chainList);
```
B0Y3R-AVA: This updates the cached set, but nothing re-runs the `enabledNetworks` setter, and that setter is the only place the always-enabled set gets unioned into `_enabledNetworks`.

`init()` is fine because the dispatch on 431 lands before `networkAvailability` on 433. But `_initChainList`'s own dispatch on 539 can land after `init()`, right? `init()` races `Promise.any([_chainListFetched.promisify(), _getCachedChainList()])`, so on a slow fetch the cached list wins and the real one turns up later.

In that window a newly always-enabled network gets decorated `isAlwaysEnabled: true` for the UI but isn't in `_enabledNetworks`, so the toggle renders locked while actually sitting off until something else runs the setter. Would `this.enabledNetworks = this.#convertNetworkAvailabilityToEnabledNetworks(this._networkAvailability)` at the end of this close it?

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -257,7 +270,7 @@ export class NetworkService implements OnLock, OnStorageReady {
 
   private set enabledNetworks(networkIds: number[]) {
     const uniqueNetworkIds = [
-      ...new Set([...NETWORKS_ENABLED_FOREVER, ...networkIds]),
+      ...new Set([...this.#alwaysEnabledNetworks, ...networkIds]),
```
B0Y3R-AVA, line 273: Not from this PR, but it matters more now: the dispatch a few lines down sends `networkIds`, not `uniqueNetworkIds`, so subscribers get the list before the always-enabled union. That gap used to be a fixed set of chains and is now whatever the API says. Worth fixing while you're in here?

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -932,6 +959,36 @@ export class NetworkService implements OnLock, OnStorageReady {
     );
   };
 
+  #applyEnablementFlags = async (
```
B0Y3R-AVA, line 964: This derives the always-enabled set from the filtered chainlist, while `#updateEnablementFromChainList` derives it from the unfiltered one. Same fact computed twice off different inputs. Looks harmless today, but is there a reason not to compute it once and reuse it?

apps/next/src/pages/Settings/components/NetworkManagement/components/NetworkToggle/NetworkToggleListItem.tsx
```diff
@@ -11,15 +11,15 @@ import { NetworkAvatar } from '../NetworkAvatar/NetworkAvatar';
...
+  isAlwaysEnabled: boolean;
   onToggle: () => void;
   onClick: () => void;
 };
 
 export const NetworkToggleListItem = ({
   network,
   isEnabled,
-  isDefault,
+  isAlwaysEnabled: isDefault,
```
B0Y3R-AVA: Any reason to keep the old name? The prop says `isAlwaysEnabled` and the body reads `isDefault`, so the next person has to hold both. Renaming the two usages below is smaller than this alias.

packages/service-worker/src/services/network/NetworkService.ts
B0Y3R-AVA: Correcting myself on the UI half of this. I said the toggle would render locked, but line 45 of `NetworkToggleListItem` is `{isDefault ? null : (<Switch/>)}`, so when `isAlwaysEnabled` is true there's no switch at all.

Which makes it worse than I described. The row shows up with no control, so it reads as force-enabled, while `getEnabledNetworks()` still leaves it out. The network sits there inert and there's nothing the user can do about it. Recompute question stands.

packages/service-worker/src/services/network/NetworkService.ts
B0Y3R-AVA: I was too soft here. Went back through it and I don't think it's harmless.

`#updateEnablementFromChainList` is subscribed to raw `_allNetworks` on line 164, which has no feature-flag filter. `isNetworkEnabledByFeatureFlags` gates Solana behind `SOLANA_SUPPORT` and 999/9999 behind `HYPERLIQUID_FEATURE`. So if Contentful marks a gated network `isAlwaysEnabled: true`, it lands in `#alwaysEnabledNetworks`, then in `_enabledNetworks` via the setter on 271, and `enableNetwork`'s early return on 305 treats it as already on. `activeNetworks` still hides it from the picker, so it ends up enabled in the service layer while invisible in the UI.

The union-regardless-of-flags part isn't new. Solana is already in `NETWORKS_ENABLED_FOREVER` and gets unioned whether the flag is on or off. What's new is that a Contentful edit can now put anything into that set, which is the same shape as the HyperEVM leak we had on mobile.

Should this one run off the feature-flag-filtered list rather than raw `_allNetworks`?

packages/common/src/utils/network/getNetworkEnablement.ts
B0Y3R-AVA: Two smaller things on the same theme.

`packages/ui/src/contexts/NetworkProvider/networkSortingFn.ts:8` still indexes `NETWORKS_ENABLED_FOREVER` directly for `promoteNetworks`, so sort order won't follow the API flag in either direction. Cosmetic, but it's a consumer that didn't get migrated. Left alone on purpose?

And one you can ignore: I went looking at the `onLock` reset on 342 to see if the Contentful-driven set got stranded after an unlock, and it's fine. `StorageService.activate` fires `onStorageReady` on every unlock, so `init()` re-dispatches the chainlist and repopulates the set before `networkAvailability` runs. Nothing to do there.

meeh0w:
@B0Y3R-AVA Always-enabled networks are now a hardcoded floor the API can only add to, so a Contentful `false` can’t make C-Chain disabled. Enablement is also gated by feature flags, the enabled-network list is recomputed when the chainlist arrives late.

B0Y3R-AVA (approved):

PR #1018, opened 2026-08-27
ci: enforce coverage ratchet in PR checks

gergelylovas:
## Summary

Adds a coverage ratchet to PR CI so unit-test coverage can only go up: PRs fail when any workspace's coverage drops below the effective baseline, defined as **max(rolling baseline from the latest `main` build, checked-in floor)**.

- **Coverage collection**: every workspace's Jest config now spreads shared settings from `src/tests/coverageConfig.cjs`, collecting coverage from **all** source files (not just files imported by tests) — so adding untested code lowers the number. Coverage only runs with `--coverage` (CI); local `yarn test` is unaffected.
- **Ratchet script** (`scripts/coverage-ratchet.mjs`, dependency-free):
  - `--check [--rolling <path>] [--tolerance 0.1]` — compares current coverage against the effective baseline, renders a per-workspace table in the job summary, exits 1 on any metric dropping >0.1pp.
  - `--update` — rewrites the checked-in floor (`coverage-baseline.json`).
  - `--export <path>` — writes the current summary (used by the main-branch workflow).
- **Rolling baseline** (`coverage-baseline.yaml`, new): every push to `main` runs `yarn test:coverage` and uploads the summary as the `coverage-rolling` artifact. Improvements on `main` raise the bar automatically, no manual updates.
- **Floor** (`coverage-baseline.json`, checked in): bounds how far coverage can leak downward through repeated just-under-tolerance drops. Raise it deliberately (e.g. quarterly) via `yarn test:coverage && yarn coverage:update`.
- **PR wiring**: unit tests move from the `lint` job into a new `test-coverage` job (total CI work unchanged), which downloads the latest rolling artifact best-effort — when it's missing (first run, expired artifact, red `main`) the check degrades to floor-only with a warning. Both jobs stay secret-free; `test-coverage` needs only `actions: read` to fetch the artifact.

New root scripts: `test:coverage`, `coverage:check`, `coverage:update`.

Baseline captured from a full green run — current line coverage per workspace: service-worker 60%, common 68%, messaging 95%, inpage 87%, ui 27%, apps/next 5%, lavamoat-rspack 5%.

## Test plan

- [x] Full suite green with coverage (`yarn test:coverage`, 4m)
- [x] Ratchet passes at baseline (exit 0) and fails on a simulated drop (exit 1)
- [x] Rolling > floor raises the bar (drop vs rolling fails); rolling < floor is ignored (floor keeps the bar); missing rolling file degrades to floor-only with a warning
- [x] `--export` output round-trips through `--check --rolling`
- [x] GitHub step summary rendering verified via `GITHUB_STEP_SUMMARY`
- [x] `yarn lint` and `yarn typecheck` pass; workflows parse as valid YAML

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/coverage-ratchet.mjs
```diff
@@ -0,0 +1,236 @@
...
+  ...Object.keys(floor),
+  ...Object.keys(rolling),
+]);
+const baseline = Object.fromEntries(
+  [...workspaceNames].map((name) => [
+    name,
+    Object.fromEntries(
+      METRICS.map((metric) => [
+        metric,
+        Math.max(floor[name]?.[metric] ?? 0, rolling[name]?.[metric] ?? 0),
```
B0Y3R-AVA, line 161: Question on the effective baseline. Since rolling only ever comes from a push to main, and the floor is the only half a PR can edit, doesn't the max mean a PR that legitimately lowers coverage has no way to unstick itself? Messaging is at functions 100, so adding one untested function lands you near 99, and dropping the floor to 99 still leaves max(99, 100) = 100. Am I missing an escape hatch somewhere? If there isn't one, would you be open to a label the workflow honors, or an explicit `--allow-regression <workspace>:<metric>` so the exception lands in the diff where someone reviews it?

scripts/coverage-ratchet.mjs
```diff
@@ -0,0 +1,236 @@
...
+    `--tolerance must be a non-negative number, got: ${toleranceArg}`,
+  );
+  process.exit(1);
+}
+
+function discoverWorkspaces() {
+  const rootPkg = JSON.parse(readFileSync(join(ROOT, 'package.json'), 'utf8'));
+  const workspaces = [];
+  for (const pattern of rootPkg.workspaces) {
+    const baseDir = join(ROOT, pattern.replace(/\/\*$/, ''));
```
B0Y3R-AVA, line 72: Do we want to lean on every workspaces entry ending in `/*`? If anyone switches the root to explicit paths to drop a workspace, baseDir becomes the package itself and readdirSync goes hunting for package.json inside src/, so discovery comes back empty. I ran that case against this script and it prints "Coverage ratchet passed" and exits 0. Worth failing outright when discovery finds nothing?

scripts/coverage-ratchet.mjs
```diff
@@ -0,0 +1,236 @@
...
+        Math.max(floor[name]?.[metric] ?? 0, rolling[name]?.[metric] ?? 0),
+      ]),
+    ),
+  ]),
+);
+const failures = [];
+const improvements = [];
+const rows = [];
+
+for (const workspace of workspaces) {
```
B0Y3R-AVA, line 170: Since this walks discovered workspaces and the baseline is only a lookup, is anything verifying we actually covered everything the baseline lists? I removed a test script from a workspace sitting in the baseline at 95% and the run stayed green without mentioning it. That feels like the direction we'd want to be loud about, especially since the "not in baseline yet" warning just below already covers the harmless direction. Would a check for baseline keys missing from discovery be worth adding?

scripts/coverage-ratchet.mjs
```diff
@@ -0,0 +1,236 @@
...
+if (mode === 'update' || mode === 'export') {
+  if (missing.length > 0) {
+    console.error(
+      `Cannot write baseline, missing coverage for: ${missing.join(', ')}.\n` +
+        'Run `yarn test:coverage` first.',
+    );
+    process.exit(1);
+  }
+  const target = mode === 'export' ? exportPath : BASELINE_PATH;
+  writeFileSync(target, JSON.stringify(current, null, 2) + '\n');
```
B0Y3R-AVA, line 125: Should `--update` refuse to move the floor down? current gets written straight over the file and floor isn't read at all in this branch, so running it on a branch where some unrelated workspace regressed quietly commits a lower floor. I tested it and 90 and 95 both became 80, exit 0, no warning. Since the improvement message actively tells people to run this, it seems easy to trip over. Per metric Math.max against the existing file when it exists? I'd leave `--export` alone given rolling should mirror main either way.

scripts/coverage-ratchet.mjs
```diff
@@ -0,0 +1,236 @@
...
+
+const floor = JSON.parse(readFileSync(BASELINE_PATH, 'utf8'));
+let rolling = {};
+let baselineLabel = 'floor';
+if (rollingPath) {
+  if (existsSync(rollingPath)) {
+    rolling = JSON.parse(readFileSync(rollingPath, 'utf8'));
+    baselineLabel = 'max(rolling, floor)';
+  } else {
+    console.warn(
```
B0Y3R-AVA, line 146: This is the piece I'd most want to be louder. Every way the rolling baseline goes missing (not on main yet, workflow renamed, 90 day expiry, red main, artifact deleted) ends in a green check with only this warn line, so the rolling half could stop working for months and nobody would know. Your own run shows the shape of it: "No successful main coverage build yet", then this WARN, then a pass. Could we emit `::warning::` so it surfaces in the run summary, and hard fail when `gh run list` did find a successful main run but the file isn't where we expect? That combination means something broke rather than not set up yet.

scripts/coverage-ratchet.mjs
gergelylovas: Can you give me an example for a PR legitimately lowering the coverage? The point of this to not let coverage drop, if the coverage percentage is allowed to be lowered by the PR it stops to be a ratchet.
I would say if something has to be merged without coverage we can bypass merge requirements.

meeh0w (approved):

PR #1020, opened 2026-08-28
chore: deprecate Keystone wallet support

gergelylovas:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-14993
Figma: https://www.figma.com/design/kksuPP4PnRIFDx9Sm3544F/Core-Extension-2026?node-id=13075-88128&t=9oYaU7TyGQGG0El5-1

Deprecates Keystone wallet support. Removes Keystone onboarding, signing and device-connection code: the `ConnectKeystoneFlow` onboarding flow, QR/USB approval overlays, Keystone UI providers, the service-worker `KeystoneService`/wallets/handlers, and all `@keystonehq/*` dependencies.

Users who already onboarded with a Keystone wallet fail gracefully: their vault still loads and balances/addresses display normally, a deprecation banner is shown on the portfolio page, and signing attempts are rejected with a clear error instead of crashing.

Screenshot 2026-08-28 at 21 43 28

## Notes

- `SecretType.Keystone`/`Keystone3Pro`, `KeystoneSecrets` and all storage schema migrations are intentionally kept so existing vaults keep decrypting and loading.
- `WalletService` now throws `KEYSTONE_UNSUPPORTED_ERROR` (`@core/common`) wherever it previously constructed a Keystone signer.
- Lavamoat `policy.json` files are untouched — they are regenerated at build time, and the stale Keystone entries are harmless.

## Testing

- `yarn lint`, `yarn typecheck` and `yarn test` (211 suites / 1753 tests) all pass.
- `yarn scanner` was run to update the en locale.

## Testing Instructions

1. Onboarding: open "Access existing wallet" — the "Add using Keystone" option is gone (Recovery Phrase and Ledger remain).
2. With a wallet that was previously onboarded via Keystone: unlock normally, balances and addresses load, and the red deprecation banner appears on the portfolio home.
3. Attempt to sign (send/swap/dApp tx) with the Keystone wallet — the request fails with "This wallet type is no longer supported".

## Media

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/service-worker/src/services/wallet/WalletService.ts
```diff
@@ -384,40 +379,7 @@ export class WalletService implements OnUnlock {
...
-        return new KeystoneWallet(
-          secrets.masterFingerprint,
-          accountIndexToUse,
-          this.keystoneService,
-          network.chainId,
-          tabId,
-          evmExtendedPubKey.key,
-          avmExtendedPubKey ? avmExtendedPubKey.key : undefined,
-        );
+        throw new Error(UNSUPPORTED_WALLET_TYPE_ERROR);
```
B0Y3R-AVA, line 382: Now that this throws `UNSUPPORTED_WALLET_TYPE_ERROR` instead of building a signer, is there a test for it anywhere? The old `signs BTC transactions correctly using BitcoinKeystoneWallet` and `signs evm tx correctly using KeystoneWallet` tests got deleted in this PR and nothing replaced them with a `toThrow` assertion, so none of these three branches is currently exercised.

apps/next/src/components/KeystoneDeprecationBanner.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC } from 'react';
+import { Trans } from 'react-i18next';
+import { MdError } from 'react-icons/md';
+import { Stack, Typography, useTheme } from '@avalabs/k2-alpine';
+import { useIsUsingKeystoneWallet } from '@core/ui';
+
+export const KeystoneDeprecationBanner: FC = () => {
```
B0Y3R-AVA, line 7: Is there a render test for this anywhere? Couldn't find one asserting it shows for Keystone/Keystone3Pro and stays hidden otherwise.

packages/service-worker/src/services/secrets/SecretsService.ts
```diff
@@ -1030,27 +1029,9 @@ export class SecretsService implements OnUnlock {
...
-          } catch (error) {
-            // Device connection not active or device not found, the user can add the X/P addresses manually later on.
-            console.error(
-              '[addAddress] Failed to get Avalanche extended public key from Keystone USB:',
-              error,
-            );
-          }
         }
+        // Keystone support is deprecated — without a stored XP xpub the
+        // missing AVM keys cannot be derived anymore.
```
B0Y3R-AVA: This used to at least try a USB re-fetch when the AVM xpub was missing. Since it's a silent no-op now with no error/log surfaced, and it's reachable from the live "add account" flow (not just the deleted reconnect UI), is that intentional? Feels like a user could end up with an account missing its X/P address and never know why.

meeh0w (commented):
Looks good to me, I agree on the one comment from @B0Y3R-AVA about the `Add account` flow, though.

packages/service-worker/src/services/secrets/SecretsService.ts
gergelylovas: good point, I've cleaned up the whole addAccount flow for keystone wallets. There is no point in keeping it.

apps/next/src/components/KeystoneDeprecationBanner.tsx
gergelylovas: We don't really have render tests for components in the repo.

packages/service-worker/src/services/wallet/WalletService.ts
gergelylovas: tests added

B0Y3R-AVA (approved):

PR #1021, opened 2026-09-01
refactor: migrate away from /tokenlist [part 1]

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-15002
Figma:

## Notes
* Migrates "Manage Tokens" list to use `/v2/tokens` endpoint instead of the hardcoded tokenlist from `NetworkService`.

## Testing
* Go to Manage Tokens and click through it. You should still be able to see all the tokens, add custom tokens (with uniqueness checked). Infinite, paginated scroll + server-side powered filtering.

## Media

https://github.com/user-attachments/assets/507bf38d-e926-4547-88e3-7b03f2b5bd5e

### DIFF
--- a/apps/next/src/hooks/useNetworkTokensSearch.ts
+++ b/apps/next/src/hooks/useNetworkTokensSearch.ts
@@ -1,6 +1,6 @@
 import { useInfiniteQuery } from '@tanstack/react-query';
 import { useDebouncedValue } from '@tanstack/react-pacer';
-import { useMemo } from 'react';
+import { useCallback, useMemo, useRef } from 'react';
 import { TokenType } from '@avalabs/vm-module-types';
 import {
   ExtensionRequest,
@@ -185,9 +185,21 @@ export const useNetworkTokensSearch = ({
     includeSpamTokens,
   ]);
 
+  // Stop rapid onEndReached callbacks from firing multiple page requests.
+  const isLoadingMoreRef = useRef(false);
+  const loadMore = useCallback(() => {
+    if (!hasNextPage || isFetching || isLoadingMoreRef.current) {
+      return;
+    }
+    isLoadingMoreRef.current = true;
+    fetchNextPage().finally(() => {
+      isLoadingMoreRef.current = false;
+    });
+  }, [hasNextPage, isFetching, fetchNextPage]);
+
   return {
     tokens,
-    fetchNextPage,
+    loadMore,
     hasNextPage: hasNextPage ?? false,
     isLoading: isFetching && !isFetchingNextPage && !data,
     isFetchingNextPage,
--- a/apps/next/src/hooks/useNetworkTokensSearch.ts
+++ b/apps/next/src/hooks/useNetworkTokensSearch.ts
@@ -1,7 +1,6 @@
 import { useInfiniteQuery } from '@tanstack/react-query';
 import { useDebouncedValue } from '@tanstack/react-pacer';
 import { useCallback, useMemo, useRef } from 'react';
-import { TokenType } from '@avalabs/vm-module-types';
 import {
   ExtensionRequest,
   FungibleTokenBalance,
@@ -51,8 +50,9 @@ const matchesKeyword = (token: FungibleTokenBalance, keyword: string) => {
   return (
     token.name.toLowerCase().includes(normalized) ||
     token.symbol.toLowerCase().includes(normalized) ||
-    (token.type === TokenType.ERC20 &&
-      token.address.toLowerCase().includes(normalized))
+    // Case-insensitive address matching to make it easier for users typing in
+    // partial address.
+    ('address' in token && token.address.toLowerCase().includes(normalized))
   );
 };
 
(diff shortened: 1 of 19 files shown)


apps/next/src/hooks/useNetworkTokensSearch.ts
```diff
@@ -0,0 +1,209 @@
...
+const matchesKeyword = (token: FungibleTokenBalance, keyword: string) => {
+  if (!keyword) {
+    return true;
+  }
+  const normalized = keyword.toLowerCase();
+  return (
+    token.name.toLowerCase().includes(normalized) ||
+    token.symbol.toLowerCase().includes(normalized) ||
+    (token.type === TokenType.ERC20 &&
+      token.address.toLowerCase().includes(normalized))
```
B0Y3R-AVA: Am I reading this right that searching an SPL token by address can never match?

`toServerSearch` sends a base58 mint to the server as `address`, the server hands the token back, `getTokenMapper` tags it SPL, and then this line only checks `address` when the type is ERC20. A name or symbol won't contain a mint address, so it gets filtered straight back out.

If that's right, every Solana address search lands on "No results found", including tokens you already hold. The `SOLANA_ADDRESS` regex up on line 30 only exists for this case, so I'm assuming it's a missed branch rather than intent. Any reason not to drop the type check and test `address` for both? Both sides get lowercased, so base58 still compares fine.

packages/service-worker/src/services/tokens/TokenManagerService.ts
```diff
@@ -27,9 +47,129 @@ export class TokenManagerService {
...
+          }
+        }
+
+        const meta = response.data?.metadata;
+        const currentPage = meta?.currentPage ?? page;
+        const totalPages = meta?.totalPages ?? currentPage;
+        if (currentPage >= totalPages) {
+          break;
+        }
+        page = currentPage + 1;
```
B0Y3R-AVA: Is there a reason to advance off the server's `currentPage` rather than the local `page`?

If the aggregator ever echoes a fixed or stale `currentPage` while `totalPages` is higher, this loop has no way out. It refetches the same page forever while `tokens` grows by `CATALOG_PAGE_LIMIT` each pass, and it's doing that inside the service worker.

I've got no evidence the endpoint actually behaves that way, so treat this as paranoia rather than a bug. But `page + 1` terminates either way and costs you nothing.

apps/next/src/hooks/useNetworkTokensSearch.ts
```diff
@@ -0,0 +1,209 @@
...
+
+// `/v2/tokens` filters on `address` (exact) and `keyword` (name/symbol, min 2
+// chars) separately; route the search box value to whichever the API accepts.
+const toServerSearch = (
+  value: string,
+): { keyword?: string; address?: string } => {
+  if (isAddressLike(value)) {
+    return { address: value };
+  }
+  return value.length >= MIN_KEYWORD_LENGTH ? { keyword: value } : {};
```
B0Y3R-AVA, line 43: Does a one character search do what you want here?

`toServerSearch` returns `{}` for both "one char" and "no search", so the two produce an identical queryKey. `enabled: false` stops a refetch, but react-query still hands back whatever pages are already cached under that key, and the memo below folds them into `searchedTokens` before filtering.

So typing a single letter filters across however far the user had scrolled, not the full catalog. The comment says the idea is to narrow held and custom client side, which isn't quite what lands. It also means results shift depending on scroll history, and every first keystroke passes through this state.

packages/service-worker/src/services/tokens/TokenManagerService.ts
meeh0w: Updated 👌

apps/next/src/hooks/useNetworkTokensSearch.ts
meeh0w: Updated

apps/next/src/hooks/useNetworkTokensSearch.ts
meeh0w: Fixed

packages/service-worker/src/services/tokens/TokenManagerService.ts
```diff
@@ -27,9 +47,145 @@ export class TokenManagerService {
...
+      return inFlight;
+    }
+
+    const request = (async () => {
+      const tokens: NetworkContractTokenWithVerified[] = [];
+      let page = 1;
+      let complete = false;
+
+      try {
+        for (;;) {
```
gergelylovas: Should we use an async generator function instead? Having an infinite loop with a break feels a bit error prone. Also with the generator function helper you can separate the fetching logic from processing the results making this function cleaner.

packages/service-worker/src/services/tokens/utils/mapApiToken.ts
```diff
@@ -0,0 +1,41 @@
...
+  address: string;
+  name: string;
+  symbol: string;
+  decimals: number;
+  isNative: boolean;
+  logoUri: string | null;
+  isVerified?: boolean | null;
+  contractType: 'ERC-20' | 'SPL' | null;
+  networkCaip2Id: string;
+};
```
gergelylovas: Should we use the type from a generated client in /api-clients/token-aggregator instead?

apps/next/src/pages/Portfolio/components/ManageTokens/components/TokenSwitchList.tsx
```diff
@@ -17,64 +24,135 @@ interface Props {
...
+  }
+  const normalizedFilter = filter.toLowerCase();
+  return tokens.filter(
+    (token) =>
+      token.name.toLowerCase().includes(normalizedFilter) ||
+      token.symbol.toLowerCase().includes(normalizedFilter) ||
+      (token.type === TokenType.ERC20 &&
+        token.address.toLowerCase().includes(normalizedFilter)),
+  );
+};
```
gergelylovas: This feels very similar to the `matchesKeyword` function in useNetworkTokenSearch. Maybe we should move it to a util and reuse if possible?

gergelylovas (dismissed):

apps/next/src/pages/Portfolio/components/ManageTokens/components/TokenSwitchList.tsx
meeh0w: Fixed in 58f64c7

packages/service-worker/src/services/tokens/utils/mapApiToken.ts
meeh0w: Fixed in 58f64c7

packages/service-worker/src/services/tokens/TokenManagerService.ts
meeh0w: Fixed in 58f64c7

bohdan-struk-avalabs (dismissed):

gergelylovas (approved):

PR #1023, opened 2026-09-02
refactor: migrate away from /tokenlist [part 2]

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-15032

Based on #1021 
Figma: n/a

## Notes
* Moves the token lookup from `network.tokens` to an API call.

## Testing
Regression-test the "Add custom token" flow


packages/service-worker/src/services/settings/events/wallet_watchAsset.ts
```diff
@@ -44,11 +44,12 @@ export class WalletWatchAssetHandler extends DAppRequestHandler {
...
 
     try {
-      const tokenAlreadyExists = await isTokenSupported(
-        tokenAddress,
-        network,
-        settings,
-      );
+      const tokenAlreadyExists =
+        isCustomTokenAdded(tokenAddress, network, settings) ||
+        (await this.tokenManagerService.isTokenAvailable(
```
B0Y3R-AVA, line 49: Worth a cache and a timeout on this one?

`wallet_watchAsset` sits in `UNRESTRICTED_METHODS` (`PermissionMiddleware.ts:86`), so a page can call it without connecting first. Each call now turns into an aggregator request with whatever address the caller picked, and `isTokenAvailable` has no cache, no in-flight dedupe and no timeout. The aggregator client also skips `applyExponentialBackOffMiddleware`, which the balance client picks up right above it in `clients.ts`.

What makes me twitch is that this used to be a local array lookup with no network cost at all. `#fetchFullCatalog` already keeps `#catalogCache` and `#catalogInFlight` against the same client, so there's a pattern to copy. The timeout is the piece I'd want most: the fail-open catch only runs once the promise settles, so nothing bounds a connection that just hangs.

meeh0w: Yup, added in 0c32b64.

gergelylovas (dismissed):

bohdan-struk-avalabs (approved):

PR #1025, opened 2026-09-02
refactor: migrate away from /tokenlist [part 3]

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-15033
Based on #1023 

## Notes
* Migrates from the `/tokenlist` endpoint to `/v2/networks`, leaving the former as a fallback.

## Testing
* Regression test the extension bootup, onboarding and all the network lists

### DIFF
--- a/packages/service-worker/src/services/network/NetworkService.ts
+++ b/packages/service-worker/src/services/network/NetworkService.ts
@@ -540,14 +540,13 @@ export class NetworkService implements OnLock, OnStorageReady {
       return v2Networks;
     }
 
-    // Report every fall back to the legacy endpoint so we can monitor
-    // if (or how often) /v2/networks is unavailable before removing the legacy path.
     Monitoring.sentryCaptureException(
-      v2Error instanceof Error
-        ? v2Error
-        : new Error('Falling back to the legacy /tokenlist endpoint'),
+      new Error('NetworkService fell back to the legacy /tokenlist endpoint'),
       Monitoring.SentryExceptionTypes.NETWORKS,
-      { attempt },
+      {
+        attempt,
+        reason: v2Error instanceof Error ? v2Error.message : String(v2Error),
+      },
     );
 
     const [legacyChainList] = await resolve(
--- a/packages/service-worker/src/services/network/NetworkService.ts
+++ b/packages/service-worker/src/services/network/NetworkService.ts
@@ -545,7 +545,11 @@ export class NetworkService implements OnLock, OnStorageReady {
       Monitoring.SentryExceptionTypes.NETWORKS,
       {
         attempt,
-        reason: v2Error instanceof Error ? v2Error.message : String(v2Error),
+        reason: v2Error
+          ? v2Error instanceof Error
+            ? v2Error.message
+            : String(v2Error)
+          : '/v2/networks returned no usable networks',
       },
     );
 
--- a/packages/service-worker/src/services/network/NetworkService.ts
+++ b/packages/service-worker/src/services/network/NetworkService.ts
@@ -52,8 +52,15 @@ import {
   getDefaultEnabledNetworkIds,
   isSyncDomain,
 } from '@core/common';
-import { isSolanaNetwork, isHyperliquidNetwork } from '@core/common';
+import {
+  isSolanaNetwork,
+  isHyperliquidNetwork,
+  Monitoring,
+} from '@core/common';
 import { GlacierService } from '../glacier/GlacierService';
+import { getAuthHeaders } from '../appcheck/utils/getAuthHeaders';
+import { getV2Networks } from '~/api-clients/token-aggregator';
+import { mapV2NetworksToChainList } from './utils/mapV2Networks';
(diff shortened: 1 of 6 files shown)


packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -543,6 +533,62 @@ export class NetworkService implements OnLock, OnStorageReady {
...
+  async #fetchChainList(attempt: number): Promise<ChainList | undefined> {
+    const [v2Networks, v2Error] = await resolve(this.#fetchNetworksFromApi());
+
+    if (v2Networks && Object.keys(v2Networks).length > 0) {
+      return v2Networks;
+    }
+
+    // Report every fall back to the legacy endpoint so we can monitor
+    // if (or how often) /v2/networks is unavailable before removing the legacy path.
+    Monitoring.sentryCaptureException(
```
B0Y3R-AVA: Will this actually give you the number you're after?

The comment says it's here to track how often `/v2/networks` is unavailable before you delete the legacy path, but I think the most common failure never arrives. A transport failure rejects as `TypeError: Failed to fetch`, `client.gen.ts` rethrows it unchanged, it satisfies `instanceof Error`, and `sharedSentryConfig.ts:13` lists `/^.*Failed to fetch$/` under `ignoreErrors`, which `initSentryForBackground` spreads in. Sentry drops it client side, so a full outage of the endpoint would read as zero.

Non-2xx does come through, since that throws a non-Error and picks up the generic Error below. So the count is really "HTTP failures only".

There's a second bite as well. Dedupe stays active (7.66 merges your `integrations` array with the defaults instead of replacing them) and it compares message and stacktrace while ignoring `extra`, so the `{ attempt }` variants collapse into one event per worker lifetime. Giving this its own error type or message would dodge both problems.

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -543,6 +533,62 @@ export class NetworkService implements OnLock, OnStorageReady {
...
+    const response = await getV2Networks<true>({
+      baseUrl: process.env.TOKEN_AGGREGATOR_SERVICE_URL,
+      throwOnError: true,
+      headers: await getAuthHeaders(),
+      query: { includeSolana: true },
+    });
+
+    const data = response.data?.data;
+
+    return data ? mapV2NetworksToChainList(data) : undefined;
```
B0Y3R-AVA, line 541: Once this lands, does `#fetchFullCatalog` become the hot path for every network?

`mapV2NetworksToChainList` doesn't set `tokens`, and `getTokensByChainId` over in part 1 short-circuits on `network.tokens` being non-empty. Its comment calls that branch interim, there "to avoid a latency regression for full-catalog consumers before the switch to /v2/networks lands". So this PR is the thing that ends the interim, and every full-catalog consumer starts paging the aggregator instead.

Two knock-ons worth a thought before this merges. The paging loop in `#fetchFullCatalog` throws on any page failure and drops the pages it already collected, and `getNetworkTokens` swallows that into `[]`, so one flaky page mid-fetch quietly empties the swap and transfer pickers until remount. And the latency that interim branch was protecting against arrives for real, at up to 1000 tokens per page per network.

Worth noting the fallback isn't equivalent either. If `/v2/networks` fails and `/tokenlist` answers, `network.tokens` comes back populated and you're on the old path. Same wallet, two different token pipelines depending on which endpoint replied.

meeh0w: Fixed in the 1st PR as this is where the whole switch of endpoints is handled: https://github.com/ava-labs/core-extension/pull/1021

packages/service-worker/src/services/network/NetworkService.ts
meeh0w: Should be fixed now

packages/service-worker/src/services/network/NetworkService.ts
```diff
@@ -543,6 +533,65 @@ export class NetworkService implements OnLock, OnStorageReady {
...
+        attempt,
+        reason: v2Error
+          ? v2Error instanceof Error
+            ? v2Error.message
+            : String(v2Error)
+          : '/v2/networks returned no usable networks',
+      },
+    );
+
+    const [legacyChainList] = await resolve(
```
gergelylovas: Do we need to keep this? What is the reason we need the legacy token list as a fallback?

gergelylovas (dismissed):

packages/service-worker/src/services/network/NetworkService.ts
meeh0w: I guess you're right, there's no real reason to keep it. Wanted to play it safe, but if there is anything wrong, we'll catch it in the blue build since blue already fetches from prod.

meeh0w: Dropped the fallback in eedf286cc

bohdan-struk-avalabs (approved):

PR #954, opened 2026-06-16
feat: wire fusion-sdk into recurring swaps

meeh0w:
# Summary
Jira ticket: https://ava-labs.atlassian.net/browse/CP-14437
Figma: _n/a_

## Notes
* ⚠️ Based on #951 

## Testing
1. Order a recurring swap (verify validation works)
2. Cancel your recurring swap

### Validation & order creation

https://github.com/user-attachments/assets/37a59b96-bc54-4725-8f4a-223c2bc4f5a2

### Cancellation

https://github.com/user-attachments/assets/7dd4150c-efb3-498b-ad37-424a0c5d8aa8

PR #887, opened 2026-04-08
chore: Pin Github Actions

GeneralZero:
## Description

Github Action Pinning

PR #16, opened 2024-08-02
fix: allow reattaching forgotten ledger device

meeh0w:
## Description
* [CP-9029](https://ava-labs.atlassian.net/browse/CP-9029)

## Testing
1. Remove your Ledger device for extension in Chrome settings: `chrome://settings/content/usbDevices`
2. Try to sign any transaction
3. You should be prompted to plug in your Ledger device, but it won't work.
4. Click `Unable to connect?`
5. In the new tab, click "Reconnect" and device selection should show up.
6. Once connected, your next transaction attempt should succeed.

## Screenshots:

Uploading Screen Recording 2024-08-02 at 22.41.58.mov…

## Checklist for the author
- [ ] I've covered new/modified business logic with Jest test cases.
- [x] I've tested the changes myself before sending it to code review and QA.

PR #1030, opened 2026-09-16
fix: batched security fixes

meeh0w:
# Summary

Jira ticket: https://ava-labs.atlassian.net/browse/CP-15044

**Related VM Modules PR:** https://github.com/ava-labs/core-vm-modules/pull/473

## Notes

`main` is fully contained in the branch (it is the merge-base). The commits in `main..fix/security-hardening` fall into three groups:

- **Group A** — polished `fix:` / `test:` changes (Aug 20–27). **Genuine, curated security hardening.**
- **Group B** — the `#954`–`#1020` PRs replayed onto the branch. Subjects match PRs already in `main` but the hashes differ. **Duplicates of old `main` — NOT security hardening.**
- **Group C** — the original security work-in-progress (Jun 18–23), most of it later re-landed and polished as Group A.

A handful of Group A changes have since been **reverted or relocated to vm-modules** — see the section at the end. Those are excluded from the Group A list below.

---

## Group A — security hardening (current)

1. Expand scientific-notation token amounts (e.g. `1.23e+21`) to plain decimals before the UI truncates the fraction, so amounts orders of magnitude apart no longer render identically.
2. Route Solana RPC by chain id (not a single `isTestnet` flag) and verify the tx blockhash belongs to the cluster being signed for (skipping durable-nonce txs). *(Now relocated to `svm-module` — see end. The extension's `#assertSolanaTxBelongsToNetwork` was removed.)*
3. Strip `signingData`/`signingRequests`/`displayData` from `ACTION_UPDATED` events sent to dApp pages, keeping only correlation fields.
4. Stop dApps seeding privileged approval fields (`actionId`, `signingData`, `tabId`) by copying only needed request fields; `addAction` rejects missing/duplicate ids; the module-handled marker is honored only if the approval controller owns the id.
5. Require a usable chain ID when approving a custom network; reject when neither `chainId` nor `caipId` is present (previously yielded `NaN`).
6. Bind signing to the active account: assert the approved signer still matches the active account (EVM/BTC + Avalanche address-map check), and reject mixed-signer batches.
7. Require manual approval when swap slippage/fees are out of range (negative, or total ≥100%) and validate `minAmountOut` is a positive bigint; disable auto-approve otherwise.
8. Cancel a connection's pending approvals when the dApp port disconnects (matched on tabId **and** domain), closing the orphaned approval window.
9. Add a clarifying comment + test cleanup for action-event scoping. *(Near no-op — the actual tabId/domain filter is introduced by the "limit events data" change in Group C.)*
10. Ignore a dApp-provided `walletId` in `avalanche_addAccount`; always use the active account's wallet.
11. Reject invalid BIP-39 recovery phrases on both import and onboarding, with a proper error type/message.
12. Warn the user on the connection approval screen when the Blockaid scan errored, distinct from "malicious".
13. Fence untrusted wallet data (token/contact/account/network names) between markers in Concierge prompts, sanitize/length-cap those values, add a trust-and-safety policy, and require user confirmation for side-effecting actions (switch account, add contact, open dApp, enable/disable network); restrict opening dApps to `https:`.
14. Reject non-numeric user amounts before parsing (`stringToBigint` now throws on non-decimal input; swap amount inputs are regex-gated).
15. Validate recipient addresses against the active network (BTC + X/P HRP mainnet-vs-testnet), so a mainnet address can't be sent to on testnet and vice-versa.
16. Escape untrusted URLs before interpolating into CSS `url(...)` via a new `toSafeCssUrl` helper (http/https only).
17. Block SSRF/private-host requests when resolving NFT metadata via a new `isSafeRemoteUrl` helper (https-only; blocks loopback/RFC1918/link-local/mDNS/IPv6).
18. Accept `unknown` (not an asserted shape) in `ensureMessageFormatIsValid` so validation actually runs on untrusted input.
19. Bound the BTC send fee rate (finite, `0 < rate ≤ 10_000`).
20. Drop the keystore `pass_hash` comparison and rely on AES-GCM auth-tag failure to signal a wrong password (removes a password-hash oracle).
21. `encodeURIComponent` the domain in the favicon lookup URL.
22. Add regression **tests** for custom-network chain-id and batch-signer checks (test-only, no runtime change).
23. Prettier formatting + `.gitignore` `.env.dev`→`.env.e2e` (chore, not a fix).

---

## Group C — original security WIP (mostly superseded by Group A)

1. First version of the active-account signing guard (later re-landed as Group A #6).
2. Remove the bogus `fs@0.0.1-security` dependency; adjust LavaMoat policies.
3. Pin GitHub Actions to commit SHAs across workflows (overlaps `main`'s own action-pinning PR).
4. Cap seedless MFA retries at 5, then clear the challenge and throw.
5. Call the message-format validation inside `signGenericMessage`. *(The related EIP-712 bool check was reverted — now covered by `evm-module`.)*
6. Fix the stale-action filter returning `{}` (discarding **all** actions) → return the accumulator. *(Real bug fix.)*
7. Introduce the actual `ACTION_UPDATED` scoping filter (tabId/domain) + its test suite. *(This is where the scoping lives, not Group A #9.)*
8. Allow `http://` + private hosts for custom RPC **only in dev**; keep HTTPS/SSRF rules in prod.
9. Large change despite a tiny message: SSRF/private-IP + HTTPS enforcement on `wallet_addEthereumChain`, RPC-header validation, custom-network chain-id/RPC check, deserialize DoS limits + type allow-list, signer-token identity check, and the dApp-scan `'error'` type. *(Message under-describes it — see notes.)*
10. Tighten the whitelisted/synced domains for production builds (dev domains only in dev).
11. Limit approval favicon/logo URLs to http/https (superseded by Group A #16 / #21).

---

## Group B — NOT security hardening (old `main`, duplicated)

All `#954`–`#1020` PRs replayed onto the branch, plus the two `Merge branch 'main'` commits. Subjects match PRs already in `main` (confirmed by patch-id: `#1015`, `#1009`, `#954`, `#956` are byte-for-byte present in `main`), but the branch carries re-committed copies with different hashes. Regular feature/fix/chore PRs (HyperCore/HyperEVM swaps, DCA swaps, Ledger, Crowdin, dep bumps, etc.) — nothing to describe as security work.

Note: the Group B "Pin Github Actions (#887)" PR and the Group C action-pinning change overlap — actions get pinned twice, from two lineages.

---

## Reverted from the extension (now handled in vm-modules)

These Group A changes were removed from the extension because the concern belongs in — and in most cases is already handled by — the per-chain modules:

1. **Access list** — the hard block was removed; `evm-module`'s approval builder now surfaces an "Includes an access list" notice on the approval screen instead of refusing to sign.
2. **EIP-712 non-boolean bool** — the extension check was removed; already enforced, more strongly, by `evm-module`'s typed-data validation.
3. **Malformed provided UTXOs** — the (dead) extension util was deleted; `avalanche-module`'s `get-provided-utxos` now fails closed instead of silently returning `[]`.
4. **BTC UTXO-value validation** — the extension seedless-signer check was removed; already enforced (against the on-chain UTXO set) by `bitcoin-module`.
5. **HVM ABI validation** — the extension check was removed; already enforced (against the chain-fetched ABI) by `hvm-module`.

6. **Solana RPC routing** — `svm-module`'s `get-provider.ts` had the same Devnet/Testnet-collapse bug (a single `isTestnet` flag routed both clusters to Devnet, affecting simulation and broadcast). It now routes by caip id across all callers (sign, sign-and-send, network fee, tx history, module provider). The extension keeps its own (already-correct) routing for its own use.
7. **Solana blockhash-vs-cluster guard** — moved into `svm-module` as a shared `assertTxBelongsToNetwork` util wired into the `sign-transaction` and `sign-and-send-transaction` handlers (with durable-nonce skipping and fail-open on decode/transport errors). It now rejects a mismatched-cluster transaction **before** the approval screen and is shared with mobile. The extension's `#assertSolanaTxBelongsToNetwork` and its test were removed.

Everything else (approval/action lifecycle, connections/permissions, messaging, UI rendering, keystore/MFA/onboarding, Concierge, swap, CI/deps) is genuinely extension-only.


packages/service-worker/src/services/network/handlers/utils/isAllowedRpcUrl.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export const isAllowedRpcUrl = (url: string): boolean => {
+  try {
+    const { protocol, hostname } = new URL(url);
+
+    if (protocol !== 'https:' && !(isDevelopment() && protocol === 'http:')) {
+      return false;
+    }
+
+    if (!isDevelopment() && isPrivateHostname(hostname)) {
```
gergelylovas: Would this limit builders to be able use Core with their local infrastructure for development? isDevelopment is only ever true when Core is build locally, if a dApp or L1 engineer wants to work on their project with Core, they would not be able to do so. Should we check for testnet mode instead?

packages/service-worker/src/services/network/handlers/wallet_addNetwork.ts
```diff
@@ -100,6 +101,34 @@ export class WalletAddNetworkHandler extends DAppRequestHandler<Params, null> {
...
+          ? Number(caipToChainId(network.caipId))
+          : undefined;
+
+      if (!chainId || Number.isNaN(chainId)) {
+        throw new Error('Network is missing a usable chain ID');
+      }
+
+      // `isValidRPCUrl` probes the dApp-supplied URL, so gate it with the same
+      // HTTPS/private-host policy as `wallet_addEthereumChain` to avoid SSRF.
+      if (!isAllowedRpcUrl(network.rpcUrl)) {
```
gergelylovas: Same comment here regarding limiting builders to use Core for their work.

packages/service-worker/src/services/network/handlers/wallet_addNetwork.ts
```diff
@@ -100,6 +101,34 @@ export class WalletAddNetworkHandler extends DAppRequestHandler<Params, null> {
     try {
       const { network } = pendingAction.displayData;
 
+      // `chainId` and `caipId` are both optional on the stored payload; we need
+      // one of them to know which chain the RPC URL must actually serve.
+      const chainId = network.chainId
```
gergelylovas: Should these checks be part of the saveCustomNetwork function in NetworkService instead? Any new usage of `saveCustomNetwork` would have to duplicate these checks to ensure compliance.

apps/next/src/components/CollapsedTokenAmount.tsx
```diff
@@ -28,6 +28,61 @@ const defaultRegularProps: TypographyProps = {
...
+ *
+ * Expand the notation into plain decimal digits first, using string arithmetic
+ * so no precision is lost.
+ */
+// The exponent is attacker-controlled: `1e+999999999` would otherwise allocate a
+// huge string via `String.repeat`. A real amount stays under this — at most ~78
+// integer digits (uint256) plus up to 255 fractional (ERC-20 `decimals` is a uint8).
+const MAX_EXPANDED_LENGTH = 512;
+
+export const expandExponentialNotation = (amount: string): string => {
```
gergelylovas: Should this be in a util instead so it can be used everywhere without importing react components? Also can we use Big.js instead, feels like Big.js -> toFixed can do a very similar transformation with much less code.

packages/common/src/utils/stringToBigint.ts
```diff
@@ -1,9 +1,22 @@
 import Big from 'big.js';
 
+const DECIMAL_NUMBER_REGEX = /^[+-]?(?:\d+(?:\.\d*)?|\.\d+)$/;
+
 export function stringToBigint(value: string, decimals: number): bigint {
-  const big = Big(value.replace(/,/gi, ''));
+  if (!DECIMAL_NUMBER_REGEX.test(value)) {
```
gergelylovas: This change can cause some regressions with numbers formatted like `1,000,000.123`. The comma replacement was in place since it caused some Big.js construction issues. Now the app will just return `0n` due to `safeStringToBigint` which can break functionality.

packages/common/src/utils/toSafeCssUrl.ts
```diff
@@ -0,0 +1,41 @@
+// Characters that would terminate a quoted `url("...")` token or the surrounding
+// style rule. Every one of them remains a valid URL once percent-encoded, so
+// escaping never breaks a legitimate image.
+const CSS_URL_BREAKOUT_CHARS = '"\'()\\{};';
+
+const percentEncode = (char: string): string =>
+  `%${char.charCodeAt(0).toString(16).padStart(2, '0')}`;
+
+const escapeForCssUrl = (value: string): string =>
```
gergelylovas: `CSS.escape` should do the same escaping and is a readily available browser API.

apps/next/src/pages/Concierge/model.ts
```diff
@@ -162,26 +162,56 @@ export const functionDeclarations: FunctionDeclaration[] = [
   },
 ];
 
+export const UNTRUSTED_DATA_OPEN = '<<<CORE_UNTRUSTED_WALLET_DATA>>>';
+export const UNTRUSTED_DATA_CLOSE = '<<<END_CORE_UNTRUSTED_WALLET_DATA>>>';
+
+export const MAX_UNTRUSTED_FIELD_LENGTH = 200;
+
+export const sanitizeUntrustedText = (value: unknown): unknown => {
```
gergelylovas: This feature is discontinued so not sure how much effort we should put into this but the arbitrary 200 character limit on the data will cause issues with networks, balances and tokens as they are a lot longer than 200 characters.

apps/next/src/pages/Fusion/hooks/useSwapFormError/lib/safeParseUserAmount.ts
```diff
@@ -1,11 +1,13 @@
 import { stringToBigint } from '@core/common';
 import { FungibleTokenBalance } from '@core/types';
 
+const DECIMAL_NUMBER_REGEX = /^\d*\.?\d*$/;
```
gergelylovas: Question. Is the decimal separator always a `.`?

packages/common/src/utils/isSafeRemoteUrl.ts
```diff
@@ -0,0 +1,45 @@
...
+    parsed = new URL(rawUrl);
+  } catch {
+    return false;
+  }
+
+  if (parsed.protocol !== 'https:') {
+    return false;
+  }
+
+  return !isPrivateHostname(parsed.hostname);
```
gergelylovas: As mentioned in some other comments. Localhost domain can be totally desired use case, outright disabling it will make Core not useful for builders. If builders build with another wallet, usually that's what they will recommend to their users as well.

packages/common/src/constants.ts
```diff
@@ -19,10 +19,7 @@ const DAPP_DEV_DOMAINS = [
   '127.0.0.1',
 ] as const;
 
-const SYNCED_DOMAINS_PRODUCTION_BUILD = [
-  CORE_WEB_DOMAIN,
-  CORE_WEB_PREVIEW_DOMAIN,
-] as const;
+const SYNCED_DOMAINS_PRODUCTION_BUILD = [CORE_WEB_DOMAIN] as const;
```
gergelylovas: Why removing the preview domains? It will make testing more difficult.

packages/common/src/constants.ts
```diff
@@ -45,11 +42,14 @@ export const KNOWN_CORE_DOMAINS = [
...
-  ...DAPP_DEV_DOMAINS,
-];
+export const WHITELISTED_DOMAINS = isProductionBuild()
+  ? [
+      CORE_WEB_DOMAIN,
+      PLAYGROUND_APP,
+      ...CORE_WEB_STAGING_DOMAINS,
+      ...KNOWN_AVACLOUD_DOMAINS,
+    ]
+  : [...KNOWN_CORE_DOMAINS, ...KNOWN_AVACLOUD_DOMAINS, ...DAPP_DEV_DOMAINS];
```
gergelylovas: Can you please clean this up a bit? Since these domain lists are like a matryoshka doll it's a bit hard to understand what domains are enabled and what the difference is exactly.

gergelylovas (changes requested):

